PR #2, opened 2018-07-19
Added failures to authentication and source connection. Will also try to make websocket urls valid

AlexBlumer:
Failures to connect and to authenticate are now automatically set instead of implicitly relying on the user to handle it. Additionally, EWSClient will now try to make the url for the websocket valid if it is the basic url of the Vantiq server.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
@@ -271,6 +271,8 @@ public class ExtensionWebSocketListener implements WebSocketListener{
         log.info("WebSocket open");
     }
 
+    private final String CONNECT_ERROR = "io.vantiq.sourcemgr.connectExtension";
+    
     /**
      * Translate the received message and pass it on to the related handler. Additionally, updates the client about
      * successful authentications and source connections.
@@ -314,6 +316,16 @@ public class ExtensionWebSocketListener implements WebSocketListener{
         else if (msg.get("op") == null) {
             log.debug("Http response received");
             if (client.isAuthed()) {
+                // Is an error message before successful connection to the target source
+                // This is most likely a failure related to a source connection request
+                if (!client.isConnected() && (Integer) msg.get("status") >= 300) {
+                    if (msg.get("body") instanceof Map && ((Map)msg.get("body")).get("code") instanceof String) {
+                        String errorCode = (String) ((Map)msg.get("body")).get("code");
+                        if (errorCode.startsWith(CONNECT_ERROR)) {
+                            client.sourceFuture.complete(false);
+                        }
+                    }
+                }
                 if (this.httpHandler != null) {
                     this.httpHandler.handleMessage(msg);
                 }
@@ -323,8 +335,11 @@ public class ExtensionWebSocketListener implements WebSocketListener{
             }
             else {
                 if ((int) msg.get("status") == 200 && !client.isAuthed()) {
-                    // TODO should this be completed on all receipts or only for success?
-                    client.authFuture.complete(true);
+                    // Forcibly setting in case an error occurred before succeeding
+                    client.authFuture.obtrudeValue(true);
+                }
+                else {
+                    client.authFuture.complete(false);
                 }
                 if (authHandler != null) {
                     this.authHandler.handleMessage(msg);
@@ -349,7 +364,6 @@ public class ExtensionWebSocketListener implements WebSocketListener{
                     }
                 }
                 else if (msg.get("op").equals(ExtensionServiceMessage.OP_QUERY)) {
-                    // note: fromMap does not set messageHeaders
                     if (this.queryHandler != null) {
                         this.queryHandler.handleMessage(msg);
                     }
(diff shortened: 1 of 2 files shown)


extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -271,6 +271,8 @@ public void onOpen(WebSocket webSocket, Response response) {
         log.info("WebSocket open");
     }
 
+    private final String CONNECT_ERROR = "io.vantiq.sourcemgr.connectExtension";
```
fhcarter: This might be better placed in ExtensionServiceMessage

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -314,6 +316,16 @@ public void onMessage(ResponseBody body) {
         else if (msg.get("op") == null) {
             log.debug("Http response received");
             if (client.isAuthed()) {
+                // Is an error message before successful connection to the target source
+                // This is most likely a failure related to a source connection request
+                if (!client.isConnected() && (Integer) msg.get("status") >= 300) {
+                    if (msg.get("body") instanceof Map && ((Map)msg.get("body")).get("code") instanceof String) {
+                        String errorCode = (String) ((Map)msg.get("body")).get("code");
+                        if (errorCode.startsWith(CONNECT_ERROR)) {
```
fhcarter: Seems like if we get a different error (don't know, but things change over time), this will still hang.  Seems like any error code should abort the same way.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -323,8 +335,11 @@ else if (msg.get("op") == null) {
             }
             else {
                 if ((int) msg.get("status") == 200 && !client.isAuthed()) {
-                    // TODO should this be completed on all receipts or only for success?
-                    client.authFuture.complete(true);
+                    // Forcibly setting in case an error occurred before succeeding
+                    client.authFuture.obtrudeValue(true);
```
fhcarter: Doesn't this return that things worked regardless of if they did?

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -361,9 +375,10 @@ else if (msg.get("op").equals(ExtensionServiceMessage.OP_QUERY)) {
                     log.warn("ExtensionServiceMessage with unknown/unexpected op '" + msg.get("op") + "'");
                 }
             }
-            else if (msg.get("op").equals(ExtensionServiceMessage.OP_CONFIGURE_EXTENSION)) {
+            else if (msg.get("op").equals(ExtensionServiceMessage.OP_CONFIGURE_EXTENSION) && client.isAuthed()) {
+                // Forcibly setting in case an error occurred before succeeding
+                client.sourceFuture.obtrudeValue(true);
```
fhcarter: ditto previous comment...

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
AlexBlumer: This and the next one were intentional. What it does is force the value to be true even if it was false before, i.e. if it tries and succeeds after failing. That way anything that saved the value updates to anything that ever called it. However, I did just check and realize that it may cause issues with some queue-like functionality I implemented for websocket->authentication->connection, so I will go back and work on that bit.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
AlexBlumer: Fixed but not committed/pushed yet.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
AlexBlumer: Will do.

AlexBlumer: Actually, without the check in the next comment this constant isn't necessary.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -314,6 +315,11 @@ public void onMessage(ResponseBody body) {
         else if (msg.get("op") == null) {
             log.debug("Http response received");
             if (client.isAuthed()) {
+                // Is an error message before successful connection to the target source
+                // This is most likely a failure related to a source connection request
+                if (!client.isConnected() && (Integer) msg.get("status") >= 300) {
+                    client.sourceFuture.complete(false);
```
fhcarter: Probably worth logging a debug message here  to note that things failed.  Can be fixed next time around (that is, no reason to hold this up).

PR #4, opened 2018-07-24
Ajb api

AlexBlumer:
Fixed concurrency, issue with validifyUrl, and can now get the source name from an EWSClient.


AlexBlumer:
Wrong target

PR #6, opened 2018-07-24
Added unit tests, fixed a bug

AlexBlumer:
Added unit tests for EWSClient and copied the old ones over for EWSListener.

Also found and fixed a copy-paste error where calling connectToSource in a synchronous way would try to send an authentication message rather than a connection request.


AlexBlumer:
Wrong target again branch again

PR #7, opened 2018-07-24
Added unit tests, fixed a bug

AlexBlumer:
Added unit tests for EWSClient and copied the old ones over for EWSListener.

Also found and fixed a copy-paste error where calling connectToSource in a synchronous way would try to send an authentication message rather than a connection request.

### DIFF
--- a/extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
+++ b/extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
@@ -1,0 +1,110 @@
+package io.vantiq.extjsdk;
+
+import okhttp3.ws.WebSocket;
+import okio.Buffer;
+import okhttp3.RequestBody;
+
+import java.io.IOException;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestExtensionWebSocketClient {
+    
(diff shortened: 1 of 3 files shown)


extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+    
+    @Test
+    public void testConnectToSource() throws InterruptedException {
+        client.authFuture.complete(true);
+        client.authSuccess.complete(null);
+        
+        client.connectToSource();
+        // do twice because the initial attempt is asynchronous, so the data will not be ready for the asserts 
+        client.connectToSource();
```
AlexBlumer: I feel like there should be a better way to wait for an asynchronous function to run, but I don't know what.

fhcarter: Well, if it's a test, Thread.sleep() works. :-)

Otherwise, the trick is usually to have connectToSource() return it's completable, then just do a .get() on it -- to force it to run.

AlexBlumer: Ok. I tried Thread.sleep() in one of tests, and it seems like it's a problem with that test alone, because sleeping works for the others. I'll look into it and try to add the fix in with the reconnection PR.

PR #9, opened 2018-07-24
Adding API for reconnection messages.

AlexBlumer:
Added handler for reconnection messages and the option to automatically resend a connection message upon receiving a reconnection request.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
@@ -388,10 +388,6 @@ public class ExtensionWebSocketListener implements WebSocketListener{
                     }
                 }
                 else if (msg.get("op").equals(ExtensionServiceMessage.OP_RECONNECT_REQUIRED)) {
-                    client.sourceReconnection();
-                    if (client.autoReconnect) {
-                        log.info("Automatically attempting to reconnect to source '" + client.getSourceName() + "'");
-                    }
                     if (this.reconnectHandler != null) {
                         this.reconnectHandler.handleMessage(msg);
                     }
@@ -399,6 +395,10 @@ public class ExtensionWebSocketListener implements WebSocketListener{
                         log.warn("Reconnect received with no handler set and no autoconnect. Can no longer "
                                 + "communicate with source '" + client.getSourceName() + "'");
                     }
+                    if (client.autoReconnect) {
+                        log.info("Automatically attempting to reconnect to source '" + client.getSourceName() + "'");
+                    }
+                    client.sourceReconnection();
                 }
                 else {
                     log.warn("ExtensionServiceMessage with unknown/unexpected op '" + msg.get("op") + "'");
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
@@ -51,6 +51,10 @@ public class ExtensionWebSocketListener implements WebSocketListener{
      */
     private Handler<Map> authHandler = null;
     /**
+     * {@link Handler} that handles reconnect messages. Set by {@link #setReconnectHandler}
+     */
+    private Handler<Map> reconnectHandler = null;
+    /**
      * {@link Handler} that handles every message received by this listener, regardless of type or situation. When set
      * this handler keeps all other handlers and related logic from firing. Setting this before a successful connection
      * will stop the connected {@link ExtensionWebSocketClient} from functioning correctly by keeping it from recording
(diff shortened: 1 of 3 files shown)


extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -370,6 +387,19 @@ else if (msg.get("op").equals(ExtensionServiceMessage.OP_QUERY)) {
...
+                        this.reconnectHandler.handleMessage(msg);
+                    }
+                    else if (!client.autoReconnect) {
+                        log.warn("Reconnect received with no handler set and no autoconnect. Can no longer "
+                                + "communicate with source '" + client.getSourceName() + "'");
+                    }
+                    if (client.autoReconnect) {
+                        log.info("Automatically attempting to reconnect to source '" + client.getSourceName() + "'");
+                    }
+                    client.sourceReconnection();
```
fhcarter: Is there a reason that client.sourceReconnection() is done outside the previous if statement.  I can see that sourceReconnection() checks for whether it should do a reconnection or not, but the name seems wrong if its really doing more than reconnection (like setting status that its now disconnected or something).  Otherwise, either I'd put the sourceReconnection() inside the if, or move the logging to inside that method.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -370,6 +387,19 @@ else if (msg.get("op").equals(ExtensionServiceMessage.OP_QUERY)) {
                         log.warn("Query received with no handler set");
                     }
                 }
+                else if (msg.get("op").equals(ExtensionServiceMessage.OP_RECONNECT_REQUIRED)) {
+                    if (this.reconnectHandler != null) {
+                        this.reconnectHandler.handleMessage(msg);
+                    }
+                    else if (!client.autoReconnect) {
```
fhcarter: The _if_ logic here is a bit twisted.  It seems to have !autoReconnect as an alternative to having a handler.  Then the 
_else if !reconnect... if reconnect..._ seems odd.  Seems like something like

`if handler call handler

if !autoReconnect then
    whine about it
else
    do it.`

would seem more appropriate.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -386,7 +416,7 @@ else if (msg.get("op").equals(ExtensionServiceMessage.OP_CONFIGURE_EXTENSION) &&
                 }
             }
             else {
-                log.warn("ExtensionServiceMessage with unknown/unexpected op '" + msg.get("op") + "'");
+                log.warn("ExtensionServiceMessage received when not connected");
```
fhcarter: (general) -- local code format does
if (...) {
    ....
} else if (...) {
   ...
} else {
    ....
}

that is, else/else-if on same line as previous closing brace.  Not required for this review, but as going through...

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -83,8 +83,7 @@ public void testAuthenticateWithUsername() throws InterruptedException {
         client.webSocketFuture.complete(true);
         
         client.authenticate(user, pass);
-        // do twice because the initial attempt is asynchronous, so the data will not be ready for the asserts 
-        client.authenticate(user, pass);  
+        Thread.sleep(100); // Wait for asynchronous action to complete
```
fhcarter: At some point in the future, I'd add some capability to spin here for a "few" iterations.  On slower machines, different environments, etc., the 100/15ms will be unpredictable and folks won't understand what to do about failed tests.  So have something that will spin up to 5 or 10 seconds then fail.  But checking every 100ms is good as that keeps the tests short under good situations.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
AlexBlumer: You're right about the naming. What it does right now is reset the Future that tells whether the source is connected, and reconnects if told to. I'll separate the two parts out for clarity.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
fhcarter: Ah, OK.  Maybe just a comment in the !autoReconnect that say something about only complaining if no handler's there.  I assume that if the caller doesn't care about reconnected, they can skip the handler.  The configHandler() will still be called after the connect?

AlexBlumer: Yep. The error message is just there for when it has no notification that it should reconnect. It will still work if the user manually calls connectToSource(), but they don't have any reason to.

PR #14, opened 2018-07-30
EWSC no longer reusable

AlexBlumer:
No more potential concurrency issues with EWSC closing then quickly reopening, because that is no longer possible.

Additionally, made validifyUrl() default to "wss://dev.vantiq.com/api/v1/wsock/websocket" when passed a null value.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -33,7 +33,7 @@ public class ExtensionWebSocketClient {
      */
     WebSocket webSocket = null;
     /**
-     * The name of ths source this client is connected to.
+     * The name of the source this client is connected to.
      */
     private String sourceName;
     /**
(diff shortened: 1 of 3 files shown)


extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -191,6 +196,11 @@ private void initializeFutures() {
      * @return      A valid url for websocket connections
      */
     protected String validifyUrl(String url) {
+        if (url == null) {
+            log.warn("No websocket url given. Using default of 'wss://dev.vantiq.com/api/v1/wsock/websocket'");
```
fhcarter: I think we should not default this.  Given that we have many international customers & will have a wider presence in different clouds as we progress, I would prefer that this be a required parameter -- that is, the caller must provide a value.

AlexBlumer:
I did so because A) this way is cleaner than the other changes I was going to make to avoid errors with quick closing and reopening of the webSocket and B) I didn't think about accidental webSocket closures.  

There is sort of a handler in the form of a CompletableFuture, but you're right, it would be better to offer a more explicit handler, or even an autoReconnect option.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
AlexBlumer: Alright. Should I just let it throw an NPE or InvalidArgument if they give a bad input?

fhcarter: Invalid argument with explanation is fine.

PR #18, opened 2018-08-01
README for sdk + a few changes and bugfixes

AlexBlumer:
* Created README
* Added a gradle wrapper for the project and removed a few unnecessary dependencies
* Added `client.initiateFullConnection()` for a one-call connection
* Removed a few default handlers and changed log levels for messages warning about missing handlers
* A few bug fixes


extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Publish
+The Publish handler is called when a message is received that was generated by Vantiq with `PUBLISH <object> TO SOURCE <source name>`. The published object can be retrieved through `<message>.getObject()`. Note that anything declared with the `USING` keyword will also be placed into the same location, e.g. `PUBLISH {"hello":"world"} TO SOURCE <sourceName> USING {"option":"one"}` will generate an object that looks like `{"hello":"world","option":"one"}`. The Publish handler receives an ExtensionServiceMessage.
+
+#### <a name="queryHandler" id="queryHandler"></a>Query
+The Query handler is called when a message is received that was generated by Vantiq with `SELECT <keys> FROM SOURCE <source name>`. Query messages expect a response, either [data](#queryResponse) or an [error](#queryError), so every Client has a default handler is created that sends back an error stating that no Query handler was set. Options specified using the `WITH` keyword are received as a Map obtained with `<message>.getObject()`.
```
fhcarter: Might be worth noting that part of the reason for the default query handler is that not every extension source will necessarily support the `query` operation.  Thus, if you're source doesn't support it, leave the handler at its default setting.

extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### <a name="listener" id="listener"></a>ExtensionWebSocketListener
+The ExtensionWebSocketListener class should only be accessed and used indirectly through handlers. If you do find a reason to access it directly, you can use `ExtensionWebSocketClient.getListener()`, but all functionality interactions with a Listener should be performed through an ExtensionWebSocketClient. If you do need to save a reference to a listener, be aware that a Client's Listener is stopped and replaced with a functionally identical Listener just before the [close handler](#closeHandler) is called. Call `listener.isStopped()` to check if this has occurred.
+
+### <a name="extSvcMsg" id="extSvcMsg"></a>ExtensionServiceMessage
+The ExtensionServiceMessage class is a helper that tightly defines what can be in a message to or from a source, and has getters for the most relevant properties.
```
fhcarter: I think the `ExtensionServiceMessage` is the class that represents the messages that are sent to/from the server.  There are some helper methods to make it easier to obtain the properties of the message, but the class is, really, more than a helper class.

fhcarter: "tightly"?  It defines it, so I don't think "tightly" adds anything.

extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### <a name="extSvcMsg" id="extSvcMsg"></a>ExtensionServiceMessage
+The ExtensionServiceMessage class is a helper that tightly defines what can be in a message to or from a source, and has getters for the most relevant properties. 
+*	`getSourceName()` returns the name of the source that sent or is receiving the message. This can be useful for identifying which Client received a message.
+*	`getObject()` returns the object that is included in many messages. 
+*	`getOp()` returns a string that states what operation is requested. Constants for each operation are provided if you wish to compare the messages.
+*	`ExtensionServiceMessage.extractReplyAddress(<message>)` returns the reply address for operations that require a reply. Currently this is only relevant for Query messages.
+
+### <a name="httpResponse" id="httpResponse"></a>Response
+The Response class is a helper that tightly defines what can be in a WebSocket message to or from the Vantiq server, and has getters for each of its properties.
```
fhcarter: ditto _tightly_ comment

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -101,6 +101,20 @@ public ExtensionWebSocketClient (String sourceName) {
         listener = new ExtensionWebSocketListener(this);
     }
 
+    /**
+     * Attempts to connect to the source using the given target url and authentication token.
+     *
+     * @param url   The url of the target Vantiq server.
+     * @param token The authentication token for the target namespace.
+     * @return      An {@link CompletableFuture} that completes as {@code true} when the connection to the source is
```
fhcarter: Probably worth noting here how one might obtain the error that caused something to fail.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -174,7 +172,7 @@ public void testQueryError() {
         assert socket.compareData("body.messageCode", errorCode);
     }
     
-    
+    int WAIT_PERIOD = 10; // Milliseconds to wait between checks on async actions
```
fhcarter: Might we worth creating an ExtensionTestBase.java or something that has this `waitUntilTrue()` method in it rather than copying.  This _technique_ is used in many of our tests to share things as well as to handle common `setup()`/`cleanup()` stuff.

extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Configuration
+The configuration handler receives the message containing a configuration document created on the Vantiq client. This message is only received upon a successful source connection, and contains the source's configuration document as a Map at `((Map) <message>.getObject()).get("config")`. The configuration handler receives an ExtensionServiceMessage.
+
+#### Reconnection
+Under certain circumstances, typically a change to the source's configuration document, the Vantiq server may disconnect from the source. The reconnection handler is called when this occurs. It is suggested that in this handler you reset any changes caused by the configuration document and then reconnect using `client.connectToSource()`. If you do not want to send the reconnect message yourself or do not wish to create your own handler, then calling `client.setAutoReconnect(true)` will tell the Client to automatically try to reconnect once any handlers are finished The reconnection handler receives an ExtensionServiceMessage.
```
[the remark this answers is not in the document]
AlexBlumer: It was a thought that occurred because in my UDP source I set/change several variables based on the configuration document, and if the port that it binds to were to change, for example, then some of the old settings would stick around and cause problems if they weren't removed here. I probably should specify the original config though.

extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+*	200 - There is some data in the message, and this is the last or only message. This can be used to complete a string of responses that used code 100.
+*	204 - There were no problems, but there is also no data to be sent. When this code is used any data sent along with it will be ignored. This can be used to complete a string of responses that used code 100.
+
+#### <a name="queryError" id="queryError"></a>Query Errors
+Query errors are sent when a Query cannot be completed successfully. To send a Query error, call `client.sendQueryError(<Query address>, <error code>, <message template>, <message parametares>)`. 
+*	The Query address is obtained the same way as for a Query response, `ExtensionServiceMessage.extractReplyAddress(<Query message>)`.
+*	The error code is a string intended to help identify where the error occurred. Possible error codes are the full class name of the class that caused the error, the class name of the Exception that caused it, or the name of the server where the error occurred.
```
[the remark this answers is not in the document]
AlexBlumer: I'll change that

extjsdk/README.md
```diff
@@ -0,0 +1,93 @@
...
+*	[ExtensionWebSocketListener](#listener) -- A listener for the Vantiq deployment. Translates Vantiq messages before passing them off to handlers. Also helps ExtensionWebSocketClient keep track of what stage in the connection it is at. 
+*	[Handler](#handler) -- A simple message handler interface, designed for simple anonymous implementation. Used by ExtensionWebSocketListener to handle various message types.
+*	[ExtensionServiceMessage](#extSvcMsg) -- A class that represents source-related messages sent from the Vantiq server.
+*	[Response](#httpResponse) -- A class that represents non source-related messages sent from the Vantiq server.
+
+## How to Include In Your Own Project
+
+### Method 1 - .jar with dependencies included
+1.	Clone this repository and navigate to <repo location>/vantiq-extension-sources.
+2.	Call `gradlew extjsdk:shadowJar` or `gradlew.bat extjsdk:shadowJar` depending on your OS.
```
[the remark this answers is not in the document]
AlexBlumer: It looks like windows doesn't need the ./ or the .bat in the commmand

extjsdk/README.md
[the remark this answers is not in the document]
fhcarter: Yeah.  Maybe there's some other section that should describe the overall flow (which can be added in a separate PR).

For the source I'm working on, I expect the config document to contain all that.  So the only local configuration will likely be "how to get to VANTIQ".  Whether that's the right way or only way, I'm not certain.  But that's the way I'm thinking about it -- the specific source's actions are dictated by the configuration document.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
AlexBlumer: This one will probably go in another PR

PR #20, opened 2018-08-02
Minor fixes

AlexBlumer:
* Fixed misspelling of `Client.initiateFullConnection()`
* Removed duplicated code

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -509,38 +509,12 @@ public class ExtensionWebSocketClient {
      * functions as false.
      */
     public void close() {
-        if (this.webSocket != null) {
-            try {
-                this.webSocket.close(1000, "Closed by client");
-            } catch (Exception e) {
-                if (!e.getMessage().equals("Socket closed")) {
-                    log.warn("Websocket has already been closed");
-                } else {
-                    log.error("Error trying to close WebSocket", e);
-                }
-            }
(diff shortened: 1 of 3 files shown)


extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -112,7 +112,7 @@ public ExtensionWebSocketClient (String sourceName) {
      * @return      An {@link CompletableFuture} that completes as {@code true} when the connection to the source is
      *              fully completed, or {@code false} when the connection fails at any point along the way.
      */
-    public CompletableFuture<Boolean> inititiateFullConnection(String url, String token) {
+    public CompletableFuture<Boolean> initiateFullConnection(String url, String token) {
```
fhcarter: I don't see any related test changes.  So that implies that this path/method isn't tested?  Is that being done a different way (some other path that includes a call here), or do we need to add a test or few to verify this method is working?

AlexBlumer: I figured that since this function tries to connect to webSocket, then auth, then source in the exact same way that other tests do, it didn't need its own individual test.

fhcarter: It's probably a good idea to ensure have at least a simple test that works.  Code drifts, changes get made only one place, interfaces change a bit, etc.  More tests is always better.

If you think about it from a code coverage perspective, you want to have tests that exercise all the lines of code that you can.

PR #35, opened 2018-08-16
Create top level readme...

fhcarter:
Also defines CODEOWNERS file for requiring reviews in master starting real soon now...


README.md
```diff
@@ -1,14 +1,333 @@
...
+}
+```
+
+Then, we would us the VANTIQ CLI to load that source implementation.
+
+```
+vantiq -s <profileName> load sourceimpls exampleImpl.json
+```
+
+where `<profileName>` is replaced by the VANTIQ profile name, and `<fileName>` identifies the file containing the definition to be loaded.
```
AlexBlumer: `exampleImpl.json` is used in the example instead of `<fileName>`. Looks like copy-paste error

fhcarter: Only one typo?  Not bad!

PR #36, opened 2018-08-16
Fix build dependency

AlexBlumer:
Creates a second jar that includes the test classes, pulls from that.


AlexBlumer:
Changed so that it checks if a socket can bind to the intended port/address, skipping if it can't.

PR #41, opened 2018-08-22
use separately installed opencv rather than openpnp

fhcarter:


fhcarter:
Looks like you're already partially thru the changes.  It's probably easier if you just take my changes to the build.gradle file and skip merging this.

fhcarter:
(Never mind -- merge conflict resolution was trivial -- ready to merge)

fhcarter:
Note:  Now expects OPENCV_LOC to be set -- for gradle and for running gradle in Eclipse/IntelliJ

This may not be the best way, but it probably works for now.

PR #45, opened 2018-08-29
Added constants for the Query response codes and a function for empty Query responses

AlexBlumer:


AlexBlumer:
I was going over the code a bit with Namir and realized that it it would be easier to read code if it used these constants rather than the relatively obtuse `200` or `204`. It's not a big deal, just something that I thought could be improved with little effort.

fhcarter:
True.  My point was that that could be done by using the constants found in either ok.http3 or apache somewhere or whatever.  That is, we don't need to define them all again.  But it's OK.  We can deal with it somewhere else.  Cause we'll need more, we'll need the error limit (300), etc.  E.g org.apache.http.HttpStatus is a pretty complete list, I think.  java.net.HttpURLConnection has many, but not all (specifically, it's missing CHUNKED, at least).

AlexBlumer:
I didn't think about using constants from other . It might be still be good to define what they mean in our specific context, but not really necessary to do so.

fhcarter:
It's OK -- it makes things clearer.  Go ahead & put it in.

PR #49, opened 2018-08-31
Gradle improvements

AlexBlumer:
* UDP, extjsdk copy licenses directly into the distribution/jar instead of into src/dist

* The fat-jar for the extjsdk no longer includes  unnecessary groovy files


AlexBlumer:
Found a few more mistakes in the object recognition readme, I'm going to push those up in a minute.

fhcarter:
That's fine.  Just include them.  I have an Appt this afternoon so I'll be offline in a few minutes.

PR #60, opened 2018-09-10
Update README.md

namirfa:
Added description of how to use the "threshold" value in config.


namirfa:
Let me know if my description of the threshold clarifies this.

PR #61, opened 2018-09-10
RTSP Network Stream Quick Fix

namirfa:
Quick fix to not throw exception if we catch a MalformedURLExcpetion. The code will then try to recreate the VideoCapture, and if this fails it will throw the appropriate exception.

### DIFF
--- a/objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
+++ b/objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
@@ -193,7 +193,7 @@ public class NetworkStreamRetriever implements ImageRetrieverInterface {
         } catch (MalformedURLException e) {
 //            throw new IllegalArgumentException(this.getClass().getCanonicalName() + ".unknownProtocol: "
 //                    + "URL specifies unknown protocol");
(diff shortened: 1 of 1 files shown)


objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
```diff
@@ -186,8 +191,10 @@ public void diagnoseConnection() throws ImageAcquisitionException{
             URL urlProtocolTest = new URL((String) camera);
             InputStream urlReadTest = urlProtocolTest.openStream();
         } catch (MalformedURLException e) {
-            throw new IllegalArgumentException(this.getClass().getCanonicalName() + ".unknownProtocol: "
-                    + "URL specifies unknown protocol");
+//            throw new IllegalArgumentException(this.getClass().getCanonicalName() + ".unknownProtocol: "
+//                    + "URL specifies unknown protocol");
+            log.error(".unknownProtocol: URL specifies unknown protocol");
```
fhcarter: Better would be something akin to

log.error(this.getClass().getCanonicalName() + ".unknownProtocol: Error parsing URL: ", e)

that will include the actual underlying error (e) in the error message

namirfa:
Made that quick change, let me know if this looks better.

PR #63, opened 2018-09-11
Handling Dropped Network Connection

namirfa:
Changed the ObjectRecognitionCore to retry connecting to the source, as opposed to simply exiting.


namirfa:
I changed the tests to reflect the new behavior o exitIfConnectionFails(), since this method used to close the threads but now does not. Essentially these tests now make sure that no matter how the connection fails, the thread won't be closed.

PR #73, opened 2018-09-27
Update README.md

namirfa:


objectRecognitionSource/README.md
```diff
@@ -186,8 +188,14 @@ Most of the options required for neuralNet are dependent on the specific impleme
...
-    1.   The value can be a number between 0 and 1 (i.e. 0.4, or 0.2, etc...)
-    2.   The value can be a number between 0 and 100 (i.e. 40, or 20, etc...)
+    1.  The value can be a number between 0 and 1 (i.e. 0.4, or 0.2, etc...)
+    2.  The value can be a number between 0 and 100 (i.e. 40, or 20, etc...)
+*   saveImage: Optional. The value can be one of the following three options:
+    1.  "local"     - This will save images to the disk (outputDir must be specified in order for this to work).
+    2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is specified.
+    3.  "both"      - This will save the images both to the disk and as documents in VANTIQ. (outputDir must be specified in order to save locally(
```
fhcarter: end of the line should be a ')' instead of "("

objectRecognitionSource/README.md
```diff
@@ -186,8 +188,14 @@ Most of the options required for neuralNet are dependent on the specific impleme
...
+*   outputDir: Optional. The directory in which images will be saved locally. Images will only be saved locally if saveImage is set to be either "local" or "vantiq".
```
fhcarter: I think they are saved locally only if ```saveImage``` is set to ```local``` or ```both```

objectRecognitionSource/README.md
```diff
@@ -186,8 +188,14 @@ Most of the options required for neuralNet are dependent on the specific impleme
...
+*   saveRate: Optional. The rate at which images will be saved (i.e. "saveRate": 3 - This will save every 3rd image that is captured). If not specified, the value will default to 1.
```
fhcarter: Might want to add that the value 1 means that every image is saved.

PR #74, opened 2018-10-11
Fixed issue with not getting most recent frame for RTSP.

namirfa:


objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
```diff
@@ -47,6 +47,7 @@
     VideoCapture   capture;
     String         camera;
     Logger         log = LoggerFactory.getLogger(this.getClass().getCanonicalName());
+    Boolean        isRTSP = false;
```
fhcarter: We might want to call this ```isPushProtocol``` or something more reflective of what it is.  In real life, I think we're differentiating between protocols that push data at us (_e.g._ ```rtsp```) vs. ones from which we have to pull (_e.g._ ```http(s)```).  If we make that differentiation, then it will also help us remember why we did this in the first place :-)

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
```diff
@@ -85,7 +95,10 @@ public ImageRetrieverResults getImage() throws ImageAcquisitionException {
         Mat matrix = new Mat();
         ImageRetrieverResults results = new ImageRetrieverResults();
         Date captureTime = new Date();
-
+        
+        if (isRTSP) {
+            capture = new VideoCapture(camera);
```
fhcarter: should we ```capture.release()``` the old one first?

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/imageRetriever/NetworkStreamRetriever.java
```diff
@@ -62,6 +63,15 @@ public void setupDataRetrieval(Map<String, ?> dataSourceConfig, ObjectRecognitio
         }
         if (dataSourceConfig.get("camera") instanceof String){
             camera = (String) dataSourceConfig.get("camera");
+            try {
+                URI httpCheck = new URI(camera);
+                if (!(httpCheck.getScheme().equals("http")) && !(httpCheck.getScheme().equals("https"))) {
```
fhcarter: On reflection, is this check enough?  Suppose we got Brett's URL (```https://.../playlist.m3u8```).  That would not be determined to be a push protocol, and I think it wouldn't fix the underlying problem.

Maybe we check the protocol in use OR if we're being directed there from some mpeg playlist (~~httpCheck~~ ```pushCheck.getPath().endsWith("m3u") || pushCheck.getPath().endsWith("m3u8")```).

There probably needs to be something more general, but that'll do for now.

PR #93, opened 2018-12-10
Quick log fixes

namirfa:
Simple changes to fix logs in tests and when retrieving OpenCV Lib. Also added a missing 'assumeTrue' in TestYoloProcessor.


objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -134,6 +134,10 @@ public void testInvalidData() {
 
     @Test
     public void testImageSavingLocal() throws ImageProcessingException {
+        
+        // Only run test with intended vantiq availability
+        assumeTrue(testAuthToken != null && testVantiqServer != null);
```
rblumer4: should have a string reason/message for this assert...

PR #96, opened 2018-12-11
Update README.md

namirfa:


fhcarter:
One thing -- you may want to add that the symptom you might see is that the it cannot open video files .

objectRecognitionSource/README.md
```diff
@@ -49,6 +49,9 @@ in-depth tutorial](https://opencv-java-tutorials.readthedocs.io/en/latest/01-ins
 describe how to install it. Once it's installed, copy the jar and (lib)opencv_java342.dll/.so/.dylib to a single
 folder, then set the environment variable OPENCV_LOC to that folder. Some features may depend on additional
 .dll/.so/.dylib files, such as FFmpeg for IP cameras.
+    *   **NOTE:** If you are using a Windows or Linux machine, you may need to add the `opencv_ffmpeg342_64.dll` file to the
```
rblumer4: windows uses .dll.  I think Linux would use .so ??

namirfa:
Fixed the update to reflect both comments.

Issue #92, opened 2018-12-09
ObjectRecognition:  Option to specify decoration


bruden01:
Currently, the object recognition source saves images with the bounding box for the detectec object rendered in the image.

The config should allow an option to save the image with or without the decoration.

That way I could decorate it on my own.  For example If I only want to decorate objects of certain types.  etc...

Issue #91, opened 2018-12-09
ObjectRecognition: can't tell what image goes with what source


bruden01:
When a message is received from the object recognition source it looks like

```
{
   "timestamp": 1544324954272,
   "results": [
      {
         "confidence": 0.87103534,
         "location": {
            "top": -13.647221,
            "left": -121.504875,
            "bottom": 447.84503,
            "right": 547.24585
         },
         "label": "person"
      }
   ],
   "dataSource": {},
   "neuralNet": {}
}
```

... when the source  config is set to save images they are stored in VANTIQ and locallly under a name like ` 2018-12-09--10-22-54.jpg`

With multiple image sources it will be difficult to know which image  goes with which source.  Ideally we should add the image name to the json object so we know which one to get if we want to display it.

bruden01:
also, it would be ideal, rather than just being written to docs if they wound up in a folder structure like

objectRecognition/<source name>/<timestamp>.jpg

namirfa:
@bruden01 
I had a few questions regarding this.

1) Is it sufficient to just add the source name to the json object, or would you like the option to add the name as a prefix/suffix to the image filename.

2) When adding the folder structure to the docs, do you simply want to add an "objectRecognition" folder, which contains all the images from all your sources.

OR

Would you want to have a global "objectRecognition" folder, and then subfolders under that directory with the name of a given source.

namirfa:
Additional features as requested by Brett:

1) Change filename to include source name, e.g. objectRecognition/mySource/2018-12-17--14-57-47.jpg

2) Add filename as part of the JSON Data sent back to VANTIQ

PR #98, opened 2018-12-13
Naf obj #92

namirfa:
Closes Issue #92 

Allows option for saving images with or without bounding boxes.

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -201,7 +201,7 @@ Most of the options required for neuralNet are dependent on the specific impleme
     2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is specified.
     3.  "both"      - This will save the images both to the disk and as documents in VANTIQ. (outputDir must be specified in order to save locally)
     
-**NOTE:** All of the following options are only necessary if the "saveImage" option has been set.
+**NOTE:** All of the following options are relevant only if the "saveImage" option has been set.
 
 *   outputDir: Optional. The directory in which images will be saved locally. Images will only be saved locally if saveImage is set to be either "local" or "both".
 *   saveRate: Optional. The rate at which images will be saved (i.e. "saveRate": 3 - This will save every 3rd image that is captured). If not specified, the value will default to 1 which saves every captured image.
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -200,6 +200,9 @@ Most of the options required for neuralNet are dependent on the specific impleme
     1.  "local"     - This will save images to the disk (outputDir must be specified in order for this to work).
     2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is specified.
     3.  "both"      - This will save the images both to the disk and as documents in VANTIQ. (outputDir must be specified in order to save locally)
+    
+**NOTE:** All of the following options are only necessary if the "saveImage" option has been set.
+
 *   outputDir: Optional. The directory in which images will be saved locally. Images will only be saved locally if saveImage is set to be either "local" or "both".
 *   saveRate: Optional. The rate at which images will be saved (i.e. "saveRate": 3 - This will save every 3rd image that is captured). If not specified, the value will default to 1 which saves every captured image.
 *   labelImage: Optional. If set to "true", images will be saved with bounding boxes and labels. If set to "false", or if not set at all, the images will be saved with no bounding boxes or labels.
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -202,6 +202,7 @@ Most of the options required for neuralNet are dependent on the specific impleme
     3.  "both"      - This will save the images both to the disk and as documents in VANTIQ. (outputDir must be specified in order to save locally)
 *   outputDir: Optional. The directory in which images will be saved locally. Images will only be saved locally if saveImage is set to be either "local" or "both".
 *   saveRate: Optional. The rate at which images will be saved (i.e. "saveRate": 3 - This will save every 3rd image that is captured). If not specified, the value will default to 1 which saves every captured image.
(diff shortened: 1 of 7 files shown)


bruden01:
@namirfa, issue #91 will be important as well since if you want to draw the boxes yourselves or display images from a particular source you'll need to know which image to use.

namirfa:
@bruden01 
Working on that one next!

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -56,20 +57,23 @@
...
+     * @param imageUtil     The instance of the ImageUtil class used to save images. Either initialized, or set to null.
+     * @param outputDir     The directory to which images will be saved.
+     * @param labelImage    The boolean flag signifying if images should be saved with or without bounding boxes. If true,
+     *                      the frames will be saved with bounding boxes, and vice versa.     
+     * @param saveRate      The rate at which images will be saved, once per every saveRate frames. Non-positive values are
+     *                      functionally equivalent to 1. If outputDir is null does nothing.
+     * @param vantiq        The Vantiq variable used to connect to the VANTIQ SDK. Either authenticated, or set to null.
      */
-    public ObjectDetector(float thresh, String graphFile, String labelFile, ImageUtil imageUtil, String outputDir, int saveRate, Vantiq vantiq) {
+    public ObjectDetector(float thresh, String graphFile, String labelFile, ImageUtil imageUtil, String outputDir, Boolean labelImage, int saveRate, Vantiq vantiq) {
```
fhcarter: Another "we should change" at the next chance.

We shouldn't use the presence of the imageUtil class as a boolean. Instead, make this detection explicit.

There seem to be a few similar things (if the VANTIQ parameter is set, use it otherwise...). 

I'm assuming there are more than one ImageUtil subclasses, one for VANTIQ, another for files?

It's a matter of style, but generally overloading meanings makes things more confusing.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -131,6 +139,129 @@ public void testInvalidData() {
...
+            }
+        } finally {
+            // delete the directory even if the test fails
+            if (d.exists()) {
+                deleteDirectory(OUTPUT_DIR);
+            }
+
+            ypImageSaver.close();
+        }
+    }
```
fhcarter: Rather than clone all this, it would be better to have some method called (something like) ```doImageSaveTest(boolean savedLabeledImage)``` that takes a parameter to figure out whether the labels should or shouldn't be saved.  Then, the value of the diff should be the value of ```saveLabeledImage```.

Then, that just gets called two ways for the two tests.  No duplication means less to change when the time comes to change it -- which it will...

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
```diff
@@ -35,18 +35,21 @@
...
  *      <li>{@code outputDir}: Optional. Config and Query. The directory in which the images (object boxes included)
  *                      will be placed. Images will be saved as
  *                      "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg"
  *                      where each value will zero-filled if necessary, e.g. "2018-08-14--06-30-22.jpg". For
  *                      non-Queries, no images will be saved if not set.
  *      <li>{@code saveImage}: Optional. Config and Query. Must be set in order to save images. Acceptable values are
  *                      "local", "vantiq", or "both".
- *      <li>{@code authToken}: Required. Config and Query. Must be set in order use VANTIQ SDK.
+ *      <li>{@code labelImage}: Optional. Config and Query. A boolean flag used to decide whether to save images with or
+ *                      or without labels.
```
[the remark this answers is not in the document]
rblumer4: to really be clear, maybe add this explicitly to the end:
  A boolean flag.  A boolean flag used to decide whether to save images with or
                  or without labels.   _Only applies if savedImages is true_.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -131,6 +139,95 @@ public void testInvalidData() {
...
+                
+                // Getting both images as byte arrays and comparing them for equality
+                byte [] originalImg = baos.toByteArray();
+                byte [] savedImg = Files.readAllBytes(d.listFiles()[0].toPath());
+                
+                // Based on labelOption, check if they are equal or not equal
+                if (labelOption) {
+                    assertFalse("Images should not be identical", Arrays.equals(originalImg, savedImg));
+                } else {
+                    assert Arrays.equals(originalImg, savedImg);
```
fhcarter: Would be better (& consistent, thus easier to read) if the "else assert" was an assertTrue with a description of the problem, analogous to the assertFalse above it.

objectRecognitionSource/README.md
```diff
@@ -200,6 +200,9 @@ Most of the options required for neuralNet are dependent on the specific impleme
     1.  "local"     - This will save images to the disk (outputDir must be specified in order for this to work).
     2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is specified.
     3.  "both"      - This will save the images both to the disk and as documents in VANTIQ. (outputDir must be specified in order to save locally)
+    
+**NOTE:** All of the following options are only necessary if the "saveImage" option has been set.
+
```
fhcarter: maybe change "only necessary" to "relevant only" -- since they are all optional, they are never "necessary".  But they are meaningless if ```saveImage``` is not true.

PR #99, opened 2018-12-13
Adding the data requested by Brett in Issue #91.

namirfa:
@bruden01 
I left a comment on the original issue regarding these changes. I addressed the baseline enhancements that you requested, but let me know if the other additions I brought up are necessary.

### DIFF
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -72,7 +72,7 @@ public class ImageUtil {
             if (fileToUpload != null) {
                 vantiq.upload(fileToUpload, 
                         "image/jpeg", 
-                        target,
+                        "objectRecognition/" + target,
                         new BaseResponseHandler() {
                             @Override public void onSuccess(Object body, Response response) {
                                 super.onSuccess(body, response);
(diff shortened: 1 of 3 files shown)


namirfa:
I will add tests to confirm the source name. 

The tests to check fetching the objects with the new names are already in place. I added "objectRecognition/" to the "lastFileName" property in NeuralNetResults, and the tests that retrieve images from VANTIQ all passed which confirms they can be retrieved.

namirfa:
Hey Fred, could you review the changes I've added?

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/NeuralNetTestBase.java
```diff
@@ -19,6 +19,7 @@
 
 public class NeuralNetTestBase extends ObjRecTestBase {
     public static final String MODEL_DIRECTORY = "src/test/resources/models";
+    public static final String SOURCE_NAME = "src";
```
fhcarter: tweaky, but having this name match where we keep our source bothers me :-).  Maybe a different name.  Im concerned that if someone confuses things or their setting, `objectRecognition/src/...` looks an awful lot like where our repo keeps code...

maybe use testSourceName as the name or something.

namirfa: Sounds good, changing this now!

Issue #103, opened 2018-12-21
Bounding Boxes in Object Recognition are incorrect


bruden01:
The bounding boxes in the object recognition source are incorrect size.  Generally, they are much too big as rendered in the image. 

This is a problem for the ford POC as the partner creating the 3d digital twin needs to use these coordinates to render the 3d twin in the correct location.  At current its unusable by the partner Prefix which is holding up the project.

This is odd though since the coordinates are smaller and in the wrong place compared to runnign it directly through darkflow.

For example for a given image we report
```
"results": [
      {
         "confidence": 0.8291698,
         "location": {
            "top": 48.742752,
            "left": -2.42628,
            "bottom": 448.82892,
            "right": 477.1719
         },
         "label": "person"
```

and darkflow reports
```
[
    {
        "confidence": 0.83,
        "bottomright": {
            "x": 1168,
            "y": 719
        },
        "label": "person",
        "topleft": {
            "x": 284,
            "y": 118
        }
    }
]
```

fhcarter:
Brett,

Part of the issue is that the values are for different images.  Yolo reports its boxes based on a 416x416 image (that's what it's tuned for).  We re-scale these for the actual image size, so it's likely ours will be larger (since the actual images are usually larger than 416x416.

That said, there's clearly something amiss as those values should never be negative (obviously).  We're working on a fix...

namirfa:
@bruden01 

I have merged a partial fix to this issue, which now returns the correctly scaled location values as part of the JSON Object.

Could you try running your image through both the Ext. Source and darkflow again, so we can see if the locations are were we would expect them to be for the differently-sized images. This should help us completely resolve the issue.

rblumer4:
Yes, Namir updated master on 2018-12-21, so you need to rebuild your _vantiq-extension-source.jar_ from the latest master code.

bruden01:
yes,  thats the one I grabbed.  I did a git pull and it says  

```
git pull
Already up to date.
```

i had updated it on the 22nd so I pulled in the change.  But I still don't see the correct results.  I will see there is some chnage as now I saw a number higher than 200.   e.g `"right": 1402.1096`  but the bounding box is not correct. I also verified this with a pb file I created that only does face recogniton.  The bounding box is not correct.  See for comparison.

The outer box is ours and the inner box when I run directly in darkflow.

[...]

PR #104, opened 2018-12-21
Changing location values returned to be the scaled values.

namirfa:
Changing the returned location values to be the same as those used to label the image.

### DIFF
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -109,12 +109,12 @@ public class ObjectDetector {
         try (Tensor<Float> normalizedImage = normalizeImage(image)) {
             Date now = new Date(); // Saves the time before
             List<Recognition> recognitions = YOLOClassifier.getInstance(threshold).classifyImage(executeYOLOGraph(normalizedImage), LABELS);
+            BufferedImage buffImage = imageUtil.createImageFromBytes(image);
             
             // Saves an image every saveRate frames
             if (imageUtil.saveImage && ++frameCount >= saveRate) {
                 String fileName = format.format(now) + ".jpg";
                 lastFilename = fileName;
-                BufferedImage buffImage = imageUtil.createImageFromBytes(image);
                 if (labelImage) {
                     buffImage = imageUtil.labelImage(buffImage, recognitions);
                 }
@@ -123,7 +123,7 @@ public class ObjectDetector {
             } else {
                 lastFilename = null;
             }
-            return returnJSON(recognitions);
+            return returnJSON(recognitions, buffImage);
         }
     }
     
@@ -150,11 +150,11 @@ public class ObjectDetector {
         try (Tensor<Float> normalizedImage = normalizeImage(image)) {
             Date now = new Date(); // Saves the time before
             List<Recognition> recognitions = YOLOClassifier.getInstance(threshold).classifyImage(executeYOLOGraph(normalizedImage), LABELS);
+            BufferedImage buffImage = imageUtil.createImageFromBytes(image);
             
             // Saves an image if requested
             if (outputDir != null || vantiq != null || (fileName != null && this.imageUtil.saveImage)) {
                 ImageUtil imageUtil = new ImageUtil();
-                //imageUtil = new ImageUtil(vantiq, outputDir);
                 imageUtil.outputDir = outputDir;
                 imageUtil.vantiq = vantiq;
                 imageUtil.sourceName = sourceName;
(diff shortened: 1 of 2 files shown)


fhcarter:
Question:  In Bret's example for the issue reported, the location values sometimes appeared to be negative.  Do we know how that happened?  It seemed like the scaling of things shouldn't ever make them negative -- they might approach zero, but shouldn't come below zero.

namirfa:
Brett's example uses the version of the Extension Source that did not return the scaled location values. I believe the negatives are because the recognized object was near the edge of the frame, and the YOLO Processor assumed that the bounding box would have originated outside of the frame.

fhcarter:
I guess I'm confused then.  In Brett's example, the numbers are way off -- not in a way that would correspond to scaling.  Unless we were "scaling" but doing completely wrong.  That is, what we returned
```
"location": {
            "top": 48.742752,
            "left": -2.42628,
            "bottom": 448.82892,
            "right": 477.1719
         },
```
vs what Darkflow reported
```
        "bottomright": {
            "x": 1168,
            "y": 719
        },
        "label": "person",
        "topleft": {
            "x": 284,
            "y": 118
        }
```

doesn't seem to be an issue of scaling.  Unless in that version, things were just borked.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -249,19 +248,21 @@ private Graph createNormalizerGraph() {
...
         	HashMap location = new HashMap();
-        	location.put("left", recognition.getLocation().getLeft());
-        	location.put("top", recognition.getLocation().getTop());
-        	location.put("right", recognition.getLocation().getRight());
-        	location.put("bottom", recognition.getLocation().getBottom());
+        	location.put("left", recognition.getScaledLocation(scaleX, scaleY).getLeft());
+            location.put("top", recognition.getScaledLocation(scaleX, scaleY).getTop());
+            location.put("right", recognition.getScaledLocation(scaleX, scaleY).getRight());
+            location.put("bottom", recognition.getScaledLocation(scaleX, scaleY).getBottom());
         	map.put("location", location);
```
fhcarter: From this, it looks like we were returning the unscaled location?  That, returning it directly from the recognition rather than scaling to the image we're saving (either locally or VANTIQ-ly)?

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -624,9 +624,9 @@ public void testQuery() throws ImageProcessingException, InterruptedException {
     }
 
     // ================================================= Helper functions =================================================
-    String imageResultsAsString = "[{\"confidence\":0.8445639, \"location\":{\"top\":254.66667, \"left\":82.87007, \"bottom\":441.9221, "
-            + "\"right\":309.81705}, \"label\":\"keyboard\"}, {\"confidence\":0.7516027, \"location\":{\"top\":88.22579, \"left\":52.281204, "
-            + "\"bottom\":330.7827, \"right\":429.71838}, \"label\":\"tvmonitor\"}]";
+    String imageResultsAsString = "[{\"confidence\":0.8445639, \"location\":{\"top\":229.5673, \"left\":99.603455, \"bottom\":398.36725, "
+            + "\"right\":372.37628}, \"label\":\"keyboard\"}, {\"confidence\":0.7516027, \"location\":{\"top\":79.53046, \"left\":62.83799, "
+            + "\"bottom\":298.18152, \"right\":516.48846}, \"label\":\"tvmonitor\"}]";
```
fhcarter: I think we should add some other tests.  We need things that have more than one entity & ones that are right on the edge, I think.  We can add those after this submission, but we do need more than one image to test.

fhcarter:
General comments: 1) PR should say something about "fixes #103".  2). Generally, we name branches with the issue on which they're working (if there's no issue, add one first).  Easier to track things & keep track of what goes in.  No need to change here, but in the future...

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
namirfa: Yes exactly, we were in fact returning the unscaled location.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
namirfa: Sounds good.

namirfa:
I didn't link this branch directly to the issue because this isn't necessarily a fix to the issue of labeling the image correctly. This was more intended to fix the issue of returning the unscaled label positions, which should give more clarity into understanding the labeling issue.

fhcarter:
ok.  Still approved as a partial fix,..

PR #106, opened 2019-01-06
Adding anchors to config options

namirfa:
Allows user to customize anchor values in Source Config. If anchors are not specified, or if they are invalid, then (new) default values will be used.


objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
```diff
@@ -125,6 +126,40 @@ private void setup(Map<String, ?> neuralNet, String sourceName, String modelDire
        } else {
            log.debug("The threshold was not specified in the config. Using default threshold value of 0.5.");
        }
+       
+       // Get anchor values if they exist
+       if (neuralNet.get("anchors") instanceof List) {
+           List tempAnchorList = (List) neuralNet.get("anchors");
+           // Checking that there are 5 anchor pairs (10 elements total)
+           if (tempAnchorList.size() != 10) {
```
fhcarter: magic numbers :-(

It would be better here to validate that the number of pairs matches YOLOClassifier.NUMBER_OF_BOUNDING_BOX(ES) * 2 (with an associated comment saying this.

In the future, we could, when we're reading the .meta file, expand this number to be variable, since (I think) it's provided in some way in the .meta file.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -120,6 +120,63 @@ public void testResults() throws ImageProcessingException {
             fail("Could not interpret json string" + e.getMessage());
         }
     }
+    
+    @Test
+    public void testValidAnchors() throws ImageProcessingException {
+        Map config = new LinkedHashMap<>();
+        YoloProcessor ypImageSaver = new YoloProcessor();
+        
+        List<Number> anchorList = Arrays.asList(1,1,1,1,1,1,1,1,1,1);
```
fhcarter: We should have a test that verifies that floating point works, too.  Based on how the code works, a single array of mixed numbers should work.

Also -- seems like we need a test that verifies that a "real" config document -- with a JSON array of either numbers or character strings (if that's accepted) will work.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/classifier/YOLOClassifier.java
```diff
@@ -19,20 +19,24 @@
  */
 public class YOLOClassifier {
     private final static float OVERLAP_THRESHOLD = 0.5f;
-    private final static double anchors[] = {1.08,1.19,  3.42,4.41,  6.63,11.38,  9.42,5.11,  16.62,10.52};
     private final static int SIZE = 13;
     private final static int MAX_RECOGNIZED_CLASSES = 24;
     private final static int MAX_RESULTS = 24;
     private final static int NUMBER_OF_BOUNDING_BOX = 5;
+    
+    private static double anchors[] = {0.57273, 0.677385, 1.87446, 2.06253, 3.33843, 5.47434, 7.88282, 3.52778, 9.77052, 9.16828};
```
rblumer4: what do these default values mean?  where did they come from?  does the array-size need to match NUMBER_OF_BOUNDING_BOX*2 ??
Add a comment explaining them...

namirfa: Chatted with Roger in person about this. Going to add an inline comment to explain what these values are (just the default values).

objectRecognitionSource/src/main/java/edu/ml/tensorflow/classifier/YOLOClassifier.java
```diff
@@ -19,20 +19,25 @@
...
     private final static float OVERLAP_THRESHOLD = 0.5f;
-    private final static double anchors[] = {1.08,1.19,  3.42,4.41,  6.63,11.38,  9.42,5.11,  16.62,10.52};
     private final static int SIZE = 13;
     private final static int MAX_RECOGNIZED_CLASSES = 24;
     private final static int MAX_RESULTS = 24;
-    private final static int NUMBER_OF_BOUNDING_BOX = 5;
+    public final static int NUMBER_OF_BOUNDING_BOX = 5;
+    
+    // Default anchor values used to properly label recognitions on original image
+    private static double anchors[] = {0.57273, 0.677385, 1.87446, 2.06253, 3.33843, 5.47434, 7.88282, 3.52778, 9.77052, 9.16828};
```
fhcarter: Maybe add a constant (private final...) called something like ANCHORS_FOR_PROVIDED_MODEL?  

Also, we should include a reference or description to the readme that outlines this a bit.  And maybe some of the links to the descriptive info that I found.  Also of interest, [object-detection](http://machinethink.net/blog/object-detection/) which talks about it... 

We should include a comment that says that ```anchor[]```  is NUMBER_OF_BOUNDING_BOX pairs of numbers, specifying the most common shapes of objects in the training data (within the YOLO 13x13 grid), and is used to construct the bounding boxes for detected objects.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
```diff
@@ -125,6 +126,34 @@ private void setup(Map<String, ?> neuralNet, String sourceName, String modelDire
...
+                       break;
+                   } else if (tempAnchorList.get(i) instanceof Integer) {
+                       anchorArray[i] = (double) ((Integer) tempAnchorList.get(i));
+                   } else {
+                       anchorArray[i] = (double) tempAnchorList.get(i);
+                   }
+               }
+           }
+       } else {
+           log.debug("Anchor values were not set, or improperly set, in the config. Anchors must be a list of "
```
fhcarter: This should probably be a ```log.warn()```

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -660,4 +743,22 @@ boolean valEquals(Object actual, Object expected) {
...
+    ExtensionServiceMessage createRealConfig() {
+        Map msg = new LinkedHashMap();
+        Map object = new LinkedHashMap();
+        Map neuralNet = new LinkedHashMap();
+        
+        neuralNet.put("pbFile", PB_FILE);
+        neuralNet.put("labelFile", LABEL_FILE);
+        neuralNet.put("outputDir", OUTPUT_DIR);
+        neuralNet.put("saveRate", SAVE_RATE);
+        neuralNet.put("anchors", Arrays.asList(1,1,1,1,1,1,1,1,1,1));
```
fhcarter: Should either have a String JSON message that we suck in via a JSON converter or something.  I want to make certain that we show what JSON we expect.

objectRecognitionSource/README.md
```diff
@@ -196,16 +197,22 @@ Most of the options required for neuralNet are dependent on the specific impleme
...
+    * (A future release will support (and encourage) supplying the `.meta` file instead of the `labelFile`. The `.meta` file will contain both the labels and the `anchors` value, as well as other data.) 
+    * Note that the anchors value for a particular model can be found in model's `.meta` file.
 *   saveImage: Optional. The value can be one of the following three options:
     1.  "local"     - This will save images to the disk (outputDir must be specified in order for this to work).
-    2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is specified.
+    2.  "vantiq"    - This will save images as documents in VANTIQ. No images will be saved locally even if outputDir is    specified.
```
fhcarter: "is    specified" --> "is specified"

spacing

Issue #107, opened 2019-01-08
ObjectRecognition:  Accept/Encourage .meta file along with .pb file


fhcarter:
We should use the information in the .meta file to properly configure the OR source.  To be used, this will require that the .meta file be provided.  We can use that in lieu of the ```labelFile``` and ```anchors``` config options (though this will remain for backward compatibility.

Over time, we can use more information from the ```.meta``` file to configure the processing and potentially handle a wider range of models.

PR #111, opened 2019-01-09
Naf obj #107

namirfa:


objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -718,6 +825,7 @@ public void testQuery() throws ImageProcessingException, InterruptedException {
             + "             \"anchors\":[0.57273, 0.677385, 1.87446, 2.06253, 3.33843, 5.47434, 7.88282, 3.52778, 9.77052, 9.16828], "
             + "             \"pbFile\": \"" + PB_FILE + "\", "
             + "             \"labelFile\": \"" + LABEL_FILE + "\", "
+            + "             \"metaFile\": \"" + META_FILE + "\", "            
             + "             \"type\": \"yolo\""
             + "         }"
             + "}";
```
fhcarter: Perhaps it happened behind the scenes, but I don't see any place where we test that the metaFile processing works.  We check that we can process it with al the overrides (above), but not where all that's provided is the pbFile & the metaFile, which is the normal case. 

I think we'll need a few more neuralNetJON-type things:  this one, one with just pb & meta file, one "the old way" (pb, labelFile, anchors) and, maybe, for good measure, one with just pbFile & labelFile, as that should work now that we've fixed the default anchor values.  I think, at this point, all should get the same answer since we cribbed the anchors from the metafile (we should check that the metaFile still has the same values). 

(These should all get the same results, since we override things in this config to be the same as the old way, there should be little chance of changing.)

fhcarter: We should update the VANTIQ project zip to be up to date -- defining only the metaFile & pbFile since that's the preferred method of operation these days.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -74,12 +81,35 @@
...
-            this.anchorArray = anchorArray;
+            // Parse meta file if it exists, and either the label file or anchor config options have not been set
+            if (metaFile != null && (labelFile == null || anchorArray == null)) {
+                parseMetaFile(metaFile);
+            }
+            // If label file exists, use it. Otherwise, use the meta file's labels.
+            if (labelFile != null) {
+                LABELS = IOUtil.readAllLinesOrExit(labelFile);
+            } else {
+                LABELS = (List<String>) metaFileMap.get("labels");
```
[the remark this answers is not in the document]
namirfa: At this point it should be non-null. If both the meta file and label file were non-null, we would not make it to this point. Otherwise, the meta file would be non-null and the label file would be null, which would trigger the previous condition and call parseMetaFile.

Let me know if I'm missing something here, or if I should still add a check.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -74,12 +81,35 @@
...
-    public ObjectDetector(float thresh, String graphFile, String labelFile, double[] anchorArray, ImageUtil imageUtil, String outputDir, 
+    public ObjectDetector(float thresh, String graphFile, String labelFile, String metaFile, double[] anchorArray, ImageUtil imageUtil, String outputDir, 
             Boolean labelImage, int saveRate, Vantiq vantiq, String sourceName) {
         try {
             GRAPH_DEF = IOUtil.readAllBytesOrExit(graphFile);
-            LABELS = IOUtil.readAllLinesOrExit(labelFile);
-            this.anchorArray = anchorArray;
+            // Parse meta file if it exists, and either the label file or anchor config options have not been set
+            if (metaFile != null && (labelFile == null || anchorArray == null)) {
+                parseMetaFile(metaFile);
```
fhcarter: Seems like we should parse the metaFile regardless of presence of labelFile or anchorArray.  No reason not too & we may use something from it in the future.

namirfa: Sure I can add that. Only reason I wasn't doing it was to save memory.

I'll change the condition to just check that the metaFile != null.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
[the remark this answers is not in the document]
fhcarter: As I read the code, it says

```
                    if (metaFile != null && (labelFile == null || anchorArray == null)) {
	                parseMetaFile(metaFile);
	            }
	            // If label file exists, use it. Otherwise, use the meta file's labels.
	            if (labelFile != null) {
	                LABELS = IOUtil.readAllLinesOrExit(labelFile);
	            } else {
	                LABELS = (List<String>) metaFileMap.get("labels");
                    }
```
it says, if the metafile is there & some other stuff is true, parse the meta file (which will create metaFileMap).

Then, regardless of whether we did that, if there's no label file, read the labelFile from the metaFileMap.

So, if we have neither a labelFile nor a metaMap, we'll blow up here.  Now, maybe someone higher up verifies that at least one of these is true, but that dependency is "far way" and pretty-well buried (down in YoloProcessor.setup()).  Personally, I'd add a check.  

Also, I noticed (didn't before) & added a comment that we should add parse any metaFile we get -- no reason not to.  Currently, I think we don't if we have a metaFile & we have both labels & anchors.  In the future, we'll use more from there, so we should verify the metaFile that they give us, I think.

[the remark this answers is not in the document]
namirfa: Ok so I just added the fix to parse the meta file whenever we get it. 

And you're exactly right, the check is in YoloProcessor.setup() which says that if we don't have either a label or a meta file, then we will throw an exception before initializing the ObjectDetector class. In any case, I'll add the check here just to be completely bullet proof.

namirfa:
Hey Fred, didn't see your comment and ended up rebasing with master. I need to clean up the README, but I believe everything else should be all set. Let me know if I messed things up.

namirfa:
Updated the example project to have a pbFile and metaFile in the config. Removed the anchors and labelFile.

objectRecognitionSource/README.md
```diff
@@ -437,7 +435,11 @@ The options are as follows. Remember to prepend "NN" when using an option in a Q
...
     [darkflow](https://github.com/thtrieu/darkflow).
-*   labelFile: Required. Config only. The labels for the model.
+*   metaFile: Required unless labelFile was supplied. Config only. A .meta file generated alongside the .pb file, that contains both the anchors and labels associated with the .pb file.
+*   labelFile: **DEPRECATED**. Required if no metaFile was supplied. Config only. The labels for the model. If both a labelFile and metaFile have been supplied, the labels from the labelFile will be used.
+*   anchors: Optional, but encouraged if a different model is used and no metaFile was provided. This value is closely tied to the model (as specified in the `neuralNet.pbFile` and `neuralNet.labelFile` configuration parameters). The `anchors` are constructed from the training data, specifying the most likely rectangles that contain objects. These are, in turn, used to define the bounding boxes for objects discovered. If not specified, the default value will be used. The default value corresponds to the correct `anchors` value for the model that is used in the build. If you use a different model, you are encouraged to supply the appropriate `anchor` values.
```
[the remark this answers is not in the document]
fhcarter: parenthetical on line 440 should now say metaFile instead of labelFile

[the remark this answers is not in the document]
fhcarter: "The labels for the model" on line 439 should be a complete sentence -- This file contains the labels for the model or something like that.

fhcarter:
Ignore last two comments. Github was seriously confused about what things looked like.  Refreshing & starting new review...

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -125,10 +232,44 @@ public void testResults() throws ImageProcessingException {
...
+            ypImageSaver.setupImageProcessing(neuralNetConfig, SOURCE_NAME, MODEL_DIRECTORY, testAuthToken, testVantiqServer);
+        } catch (Exception e) {
+            fail("Should not fail with valid config.");
+        }
+        
+        // Config with label file and pb file
+        msg = createRealConfig(neuralNetJSON4);
+        config = (Map) msg.getObject();
+        neuralNetConfig = (Map) config.get("neuralNet");
+
```
fhcarter: These tests only accept that we correctly set internal variables & classes.  No actual processing is done.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
fhcarter: We are still not testing that things work.  That's why the test results never change.

The method classSetup() creates a YOLO processor that's initialized with the model & label file.  That's all. testResults() is the only test that checks any output.  So that's the only test that checks anything.  There is no test that checks metaFile, anchors setting, etc. as far as I can tell.

All of the configurations in testRealJSONConfig() should run using the test image (or more) to verify that they work the way we expect they do.  The easiest way to do this is to create a method that does what testResults() does, but takes the YoloProcessor as an argument.  

The testResults() calls that method using ypJson (I don't know why it has that name since it actually never touches JSON!).

Then, in testRealJSONConfig(), it can call that method after each configuration.

Until we really test what we're doing, we don't have much confidence that we have things right.  More tests is always better.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -711,21 +856,64 @@ public void testQuery() throws ImageProcessingException, InterruptedException {
...
+          + "         }"
+          + "}";
 
     List<Map> getExpectedResults() throws JsonParseException, JsonMappingException, IOException {
         ObjectMapper m = new ObjectMapper();
         return m.readValue(imageResultsAsString, List.class);
     }
+    
+    void verifyProcessing(YoloProcessor ypImageSaver) throws ImageProcessingException {
+        NeuralNetResults results = ypJson.processImage(getTestImage());
```
fhcarter: Umm, this ignores the input parameter.  So still doesn't test things...

namirfa: Oh man I'm sorry that was a pretty dumb mistake. I'm fixing it now.

namirfa:
Fixed, and all the tests pass/build completes successfully.

Issue #109, opened 2019-01-09
Adding an option to save pictures only without doing object recognition


liu-xiao-guo:
Currently, there is a need in customer using object recognition extension source to save webcam pictures without the need to do object recoginition. It would be good to add a configuration to do that so that the customer can use the same source to capture webcam pictures. 

Thanks

PR #120, opened 2019-01-21
Naf obj #109

namirfa:
Fixes Issue #109 .

Adds feature that saves images as captured from the camera, without doing **any** neural net processing beforehand. Images can be saved to VANTIQ, locally, or both. This option is available for both query and polling sources.

Feature is built as a new Neural Net Interface implementation, and can be accessed by setting the source configuration's neural net type to "none". All changes are explained in more detail in the README.

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -39,7 +39,7 @@ Additionally, an example VANTIQ project named *objRecExample.zip* can be found i
     *   [YOLO Processor](#yoloNet) -- An implementation of the [You Only Look Once](https://pjreddie.com/darknet/yolo/)
         (YOLO) object detection software using Java Tensorflow.
     * [YOLO Models Available](#modelFiles)
-    * [No Processor](#noNet) -- An implementation of the interface used only to **save** images. This implementation does 
+    * [No Processor](#noNet) -- An implementation of the interface used just to **save** images. This implementation does 
     **no** image processing.
 *   [ImageRetrieverResults](#msgFormat) -- A class that holds the data passed back by image retriever implementations.
 *   [ImageRetrieverInterface](#retrieveInterface) -- An interface that allows different image retrieval mechanisms to be
@@ -506,17 +506,17 @@ Previously, `yolo.pb` file was downloaded as part of the build, but the `coco.na
 
 ### <a name="noNet" id="noNet"></a>No Processor
 
-The purpose of this implementation of the Neural Net Interface is to save images without doing any processing. This allows a 
-user to capture frames from their video or image source, and save them with VANTIQ, locally, or both. There is absolutely **no 
-processing** done to the captured frames, they are saved as is.
+The purpose of this implementation of the Neural Net Interface is to save images without doing any neural net processing. This allows a 
+user to capture frames from their video or image source, and save them with VANTIQ, locally, or both. There is **no 
+processing** done to the captured frames, they are saved as captured from the camera.
 
 The Extension Source can be setup for polling, for queries, or for both.
 
 *   outputDir: Optional. Config and Query. The directory in which the images (object boxes included) will be placed.
     Images will be saved as "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg"
     where each value will zero-filled if necessary, e.g. "2018-08-14--06-30-22.jpg". If pollTime is set under 1000 (1 
-    second), images with the same name will be followed by a count in parentheses, e.g. "2018-08-14--06-30-22(1).jpg", "2018-
-    08-14--06-30-22(2).jpg".
+    second), images with the same name, (*i.e.* captured in the same second), will be followed by a count in parentheses, e.g. 
+    "2018-08-14--06-30-22(1).jpg", "2018-08-14--06-30-22(2).jpg".
 *   fileName: Optional. Query only. The name of the file that will be saved. Defaults to
     "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg" if not set. As stated above, 
     redundant filenames will be followed by a count in parentheses.
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -39,6 +39,8 @@ Additionally, an example VANTIQ project named *objRecExample.zip* can be found i
     *   [YOLO Processor](#yoloNet) -- An implementation of the [You Only Look Once](https://pjreddie.com/darknet/yolo/)
         (YOLO) object detection software using Java Tensorflow.
     * [YOLO Models Available](#modelFiles)
+    * [No Processor](#noNet) -- An implementation of the interface used only to **save** images. This implementation does 
+    **no** image processing.
 *   [ImageRetrieverResults](#msgFormat) -- A class that holds the data passed back by image retriever implementations.
 *   [ImageRetrieverInterface](#retrieveInterface) -- An interface that allows different image retrieval mechanisms to be
         more easily integrated without changes to the rest of the code.
@@ -199,7 +201,7 @@ configuration options](#yoloNet). The ones that are the same across all implemen
 *   type: Optional. Can be one of three situations
     1.  The fully qualified class name of an implementation of NeuralNetInterface, e.g.
         "io.vantiq.extsrc.objectRecognition.neuralNet.YoloProcessor".
-    2.  The short name of one of the standard implementations, i.e. one of "[yolo](#yoloNet)" or
+    2.  The short name of one of the standard implementations, i.e. one of "[yolo](#yoloNet)", "[none](#noNet)" or
         "[default](#defaultNet)".
     3.  Empty, in which case the program will try to find an implementation with the name "DefaultProcessor" in
         the `io.vantiq.objectRecognition.neuralNet` package. This implementation is not provided, and must be written by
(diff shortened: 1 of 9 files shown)


objectRecognitionSource/README.md
```diff
@@ -39,6 +39,8 @@ Additionally, an example VANTIQ project named *objRecExample.zip* can be found i
     *   [YOLO Processor](#yoloNet) -- An implementation of the [You Only Look Once](https://pjreddie.com/darknet/yolo/)
         (YOLO) object detection software using Java Tensorflow.
     * [YOLO Models Available](#modelFiles)
+    * [No Processor](#noNet) -- An implementation of the interface used only to **save** images. This implementation does 
+    **no** image processing.
```
fhcarter: maybe "... use to just **save** images."

(my concerns is that "used only to save images" makes it sound like this is used to save images when saving them.  Just trying to reduce confusion.

objectRecognitionSource/README.md
```diff
@@ -504,6 +504,24 @@ These files are now fetched from an Amazon S3-based maven repository, and will b
 
 Previously, `yolo.pb` file was downloaded as part of the build, but the `coco.names` file was stored in Git.  That file is still in git (but likely to be removed soon).  We now obtain all the model files from the same place -- so none of this information will be in Git.  Keeping the information together is a safer way to go about things.
 
+### <a name="noNet" id="noNet"></a>No Processor
+
+The purpose of this implementation of the Neural Net Interface is to save images without doing any processing. This allows a
```
fhcarter: "... without doing any neural net processing"

objectRecognitionSource/README.md
```diff
@@ -462,9 +464,7 @@ The options are as follows. Remember to prepend "NN" when using an option in a Q
     * Example: `"anchors": [0.57273, 0.677385, 1.87446, 2.06253, 3.33843, 5.47434, 7.88282, 3.52778, 9.77052, 9.16828]`
 *   outputDir: Optional. Config and Query. The directory in which the images (object boxes included) will be placed.
     Images will be saved as "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg"
-    where each value will zero-filled if necessary, e.g. "2018-08-14--06-30-22.jpg". For non-Queries, no images will
-    be saved if not set. For Queries, either this must be set in the Query, or this must be set in the config and
-    fileName must be set in the Query for images to be saved.
+    where each value will zero-filled if necessary, e.g. "2018-08-14--06-30-22.jpg".
```
fhcarter: Are the other things (queries vs. non-queries) no longer true?

I guess this is described in more detail below.  Maybe add a link here that instructs the user that there's more detail elsewhere.

I suspect that it's worth mentioning that duplicates may have the counter.  Then the "see ... for more details" or something like that.

objectRecognitionSource/README.md
```diff
@@ -504,6 +504,24 @@ These files are now fetched from an Amazon S3-based maven repository, and will b
 
 Previously, `yolo.pb` file was downloaded as part of the build, but the `coco.names` file was stored in Git.  That file is still in git (but likely to be removed soon).  We now obtain all the model files from the same place -- so none of this information will be in Git.  Keeping the information together is a safer way to go about things.
 
+### <a name="noNet" id="noNet"></a>No Processor
+
+The purpose of this implementation of the Neural Net Interface is to save images without doing any processing. This allows a 
+user to capture frames from their video or image source, and save them with VANTIQ, locally, or both. There is absolutely **no
```
fhcarter: dump "absolutely"

objectRecognitionSource/README.md
```diff
@@ -504,6 +504,24 @@ These files are now fetched from an Amazon S3-based maven repository, and will b
 
 Previously, `yolo.pb` file was downloaded as part of the build, but the `coco.names` file was stored in Git.  That file is still in git (but likely to be removed soon).  We now obtain all the model files from the same place -- so none of this information will be in Git.  Keeping the information together is a safer way to go about things.
 
+### <a name="noNet" id="noNet"></a>No Processor
+
+The purpose of this implementation of the Neural Net Interface is to save images without doing any processing. This allows a 
+user to capture frames from their video or image source, and save them with VANTIQ, locally, or both. There is absolutely **no 
+processing** done to the captured frames, they are saved as is.
```
fhcarter: "... they are saved as captured from the camera"

objectRecognitionSource/README.md
```diff
@@ -504,6 +504,24 @@ These files are now fetched from an Amazon S3-based maven repository, and will b
...
+
+The Extension Source can be setup for polling, for queries, or for both.
+
+*   outputDir: Optional. Config and Query. The directory in which the images (object boxes included) will be placed.
+    Images will be saved as "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg"
+    where each value will zero-filled if necessary, e.g. "2018-08-14--06-30-22.jpg". If pollTime is set under 1000 (1 
+    second), images with the same name will be followed by a count in parentheses, e.g. "2018-08-14--06-30-22(1).jpg", "2018-
```
fhcarter: "... images with the same name (_i.e._ captured in the same second)..."

Is this also true (sort of a redundant question with another, below) if the user names the images with a constant? That is, I don't think we look at the save rate -- we just say, "oh, this file already exists, I'll work around that."

Should there be some parameter that allows the user to just stomp on the file rather than append the counter? (can be separate issue)

objectRecognitionSource/README.md
```diff
@@ -504,6 +504,24 @@ These files are now fetched from an Amazon S3-based maven repository, and will b
...
+    08-14--06-30-22(2).jpg".
+*   fileName: Optional. Query only. The name of the file that will be saved. Defaults to
+    "&lt;year&gt;-&lt;month&gt;-&lt;day&gt;--&lt;hour&gt;-&lt;minute&gt;-&lt;second&gt;.jpg" if not set. As stated above, 
+    redundant filenames will be followed by a count in parentheses.
```
fhcarter: Presumably, if I set the filename to "fred", I will simply get a million files named fred, fred(1), ..., fred(1000000) or something like that?

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestNoProcessor.java
```diff
@@ -0,0 +1,502 @@
...
+    static final String CONFIG_JSON_LOCATION = "src/test/resources/";
+
+    static NoProcessor noProcessor;
+    static Vantiq vantiq;
+    static VantiqResponse vantiqResponse;
+
+    static List<String> vantiqSavedFiles = new ArrayList<>();
+
+    static final String timestampPattern = "\\d{4}-\\d{2}-\\d{2}--\\d{2}-\\d{2}-\\d{2}\\.jpg";
+    static final String sameTimestampPattern = "\\d{4}-\\d{2}-\\d{2}--\\d{2}-\\d{2}-\\d{2}\\(\\d{1}\\)\\.jpg";
```
fhcarter: Shouldn't `\\(\\d{1}\\)` be `\\(\\d+\\)` or something like that (that is -- take any number in there, not just single digit ones)

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -350,14 +353,17 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
-               // Send the normal message as the response if requested, otherwise just send the data 
-               if (request.get("sendFullResponse") instanceof Boolean && (Boolean) request.get("sendFullResponse")) {
-                   Map response = createMapFromResults(imageResults, results);
-                   client.sendQueryResponse(200, replyAddress, response);
-               } else if (results.getResults().isEmpty()) {
-                   client.sendQueryResponse(204, replyAddress, new LinkedHashMap<>());
-               } else {
-                   client.sendQueryResponse(200, replyAddress, results.getResults().toArray(new Map[0]));
+               // Don't send any data if results are null (for NoProcessor)
+               if (results != null) {
```
fhcarter: Are there cases (_e.g._ failure) where results might be null?  Should we check for the NoProcessor processor rather than results == null?

If these are truly query responses, then doing nothing with processing == none is wrong.  We should still send a noData response in that case.  I think that the query in this case is used to manually poll the camera.  In that case, a query must still respond -- it'll just return no data having (presumably) saved the image as requested.  

We should definitely have a test for queries when the noProcessing option is selected.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestNoProcessor.java
```diff
@@ -0,0 +1,502 @@
...
+            if (d.exists()) {
+                deleteDirectory(OUTPUT_DIR);
+            }
+
+            npProcessor.close();
+        }
+    }
+
+    @Test
+    public void testQuery() throws ImageProcessingException, InterruptedException {
```
fhcarter: If I'm reading this correctly, this is testing the internal saving.  I think we need something that verifies that the overall source handles queries correctly when no-processing is selected.  That is, if I write a VAIL procedure that does a select against this source (which can be used, IIRC, to poll the camera, we need to ensure that the select doesn't time out/error.  It should get back a no-content which will be (correctly) interpreted as "no rows" I think.  I'm concerned because the section above seemed to simply skip returning with there was no result.   Generally, that's not good because the overall protocol from the server side is to send the query then await the result.  Failure to get it times out, but it takes a while.  And that's now how the source should behave.

(If I've misread things, please correct me!)

namirfa:
Added fixes to the review comments. Only thing I did not address was this comment:

```
Are the other things (queries vs. non-queries) no longer true?

I guess this is described in more detail below. Maybe add a link here that instructs the user that there's more detail elsewhere.

I suspect that it's worth mentioning that duplicates may have the counter. Then the "see ... for more details" or something like that.
```


That is the only change to queries that I am aware of. Additionally, the feature to add the parenthetical counter to the filename does not apply to the normal YOLOProcessor, that is unique to the NoProcessor.

PR #121, opened 2019-01-21
Make GRID_SIZE equal to FRAME_SIZE/32

namirfa:
* Changed names of "SIZE" values in YOLOClassifier and ObjectDetector to be GRID_SIZE and FRAME_SIZE respectively.
* Made GRID_SIZE equal to dividing FRAME_SIZE by 32.

### DIFF
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/Config.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/Config.java
@@ -5,6 +5,6 @@ package edu.ml.tensorflow;
  */
 public interface Config {
     // Params used for image processing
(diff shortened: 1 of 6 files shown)


objectRecognitionSource/src/main/java/edu/ml/tensorflow/Config.java
```diff
@@ -5,6 +5,6 @@
  */
 public interface Config {
     // Params used for image processing
-    public int SIZE = 416; // Edited so that it can be accessed from anywhere
+    public int FRAME_SIZE = 416; // Edited so that it can be accessed from anywhere
```
fhcarter: README needs update since we tell folks to change Config.SIZE to match their model

namirfa:
Added README fix.

Issue #122, opened 2019-01-22
Object Recognition: Extract all relevant data from .meta file, including input image size


namirfa:
Increase the meta file parsing to also retrieve the input image size if available. This will replace the hardcoded values we currently use, (which will be switched to default values).

Additionally, look for any other useful information in the meta file that can be extracted, in order to remove any other hard-coded aspects of the Object Recognition Source.

PR #125, opened 2019-01-23
Naf obj #122

namirfa:
Fixes Issue #122 .

Adds the ability to retrieve the input frame size from the .meta file. If no meta file is provided, then the standard 416 default value is used. For backwards compatibility, the user can still override both the default and the meta file's input frame size by editing the edu.ml.tensorflow.Config.FRAME_SIZE value.

I was unable to write automated tests that check the behavior when the default value is overridden. I was also unable to add tests that demonstrate the behavior if the meta file has invalid height/width values. These were both manually tested and worked as expected.

### DIFF
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -43,7 +43,7 @@ public class ObjectDetector {
     private static SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd--HH-mm-ss");
     
     // Getting meta config options for YOLO Processor
-    private MetaBasedConfig metaConfigOptions = new MetaBasedConfig();
+    public MetaBasedConfig metaConfigOptions = new MetaBasedConfig();
     private final float MEAN = metaConfigOptions.mean;
     private int frameSize;
     
@@ -95,7 +95,7 @@ public class ObjectDetector {
                 int frameHeight = (int) ((Map<String,Object>) metaFileMap.get("net")).get("height");
                 int frameWidth = (int) ((Map<String,Object>) metaFileMap.get("net")).get("width");
                 // Check that meta file has appropriate frame size, and that user has not overwritten frame size in Config
-                if (frameHeight == frameWidth && frameHeight % 32 == 0 && metaConfigOptions.useMeta) {
+                if (frameHeight == frameWidth && frameHeight % 32 == 0 && metaConfigOptions.useMetaIfAvailable) {
                     // Set config value to the .meta file's frame size
                     metaConfigOptions.frameSize = frameHeight;
                 }
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -30,6 +30,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static edu.ml.tensorflow.Config.MEAN;
+
 /**
  * ObjectDetector class to detect objects using pre-trained models with TensorFlow Java API.
  */
@@ -44,7 +46,6 @@ public class ObjectDetector {
     
     // Getting meta config options for YOLO Processor
     public MetaBasedConfig metaConfigOptions = new MetaBasedConfig();
-    private final float MEAN = metaConfigOptions.mean;
     private int frameSize;
     
     private ImageUtil imageUtil;
@@ -377,5 +378,6 @@ public class ObjectDetector {
     @Override
     protected void finalize() {
         close();
+        YOLOClassifier.getInstance(threshold, anchorArray, frameSize).close();
     }
 }
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -329,7 +329,7 @@ public class ObjectDetector {
         	map.put("confidence", recognition.getConfidence());
         	
         	float scaleX = (float) buffImage.getWidth() / (float) frameSize;
-            float scaleY = (float) buffImage.getHeight() / (float) frameSize;
+        	float scaleY = (float) buffImage.getHeight() / (float) frameSize;
         	
         	HashMap location = new HashMap();
         	location.put("left", recognition.getScaledLocation(scaleX, scaleY).getLeft());
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -30,9 +30,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static edu.ml.tensorflow.Config.MEAN;
-import static edu.ml.tensorflow.Config.FRAME_SIZE;
-
 /**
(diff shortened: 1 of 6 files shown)


namirfa:
Additionally, I did not find any other obvious values that we can pull from the .meta file. I will double check now to see if there is anything else worth adding here.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -45,6 +42,11 @@
     // "year-month-date-hour-minute-seconds"
     private static SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd--HH-mm-ss");
     
+    // Getting meta config options for YOLO Processor
+    public MetaBasedConfig metaConfigOptions = new MetaBasedConfig();
+    private final float MEAN = metaConfigOptions.mean;
```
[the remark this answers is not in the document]
namirfa: Just getting a chance to read this. So instead here, I should just leave the original MEAN as a constant taken from the Config file (how it was before), and not set it as part of the MetaBasedConfig.

Let me know if that seems about right.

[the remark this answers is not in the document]
fhcarter: @namirfa Yes, I think so.  I don't see any MEAN/mean/Mean/etc/ in any of the `.meta` or `.cfg` files, so I don't think we're gathering it from anywhere.  It's just some number that's meaningful.  Somehow?

I also looked for variations on `scale` and they don't exist either.  From some googling, I found some hints [here](https://github.com/thtrieu/darkflow/issues/38).  They appear to say that the 255 is used for normalizing the images -- I suspect that it has to do with the scaling of colors or B&W it's using.  The hint (link above) referred to that scaling number being in the range (0, 1) when running on the android version of things, so I that the 255 has to do with the number of shades involved or or something else referring to greyscale issues or something (but that's really more speculation).  It seems to be structural part of YOLO v2 (at least) that we probably shouldn't change.  It might make sense to change the name to NORMALIZATION_SCALE or something so we don't have to re-derive it each time :-).  Or maybe just add a comment to the Config file that says what it's for.  

If you read thru the graph builder & use thereof in `ObjectDetector.createNormalizerGraph(), it's decoding the JPEG into some matrix (I think), and then dividing that whole matrix (or, really, TensorFlow `Output` object) by the 255/Mean/Scale value.  So, I think it's taking the values and normalizing them into an 8-bit color/B&W scale.  I don't really think we want it to be adjustable -- and if someone wants to, they can change the Config.java file.  

So, bottom line.  Maybe I've convinced myself that it'll work as is -- pass it straight through from Config -- Don't try & read it from `.meta` (which I don't think you're doing).  Whether it has to go via `MetaBasedConfig` isn't clear.  Probably so to keep it clean, but please add a comment that says that this is not read from the `.meta` file.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/classifier/YOLOClassifier.java
```diff
@@ -46,7 +46,8 @@ public static YOLOClassifier getInstance(float thresh, double[] anchorArray, int
                 anchors = anchorArray;
             }
         }
-
+        System.out.println("Grid Size: " + gridSize);
+        System.out.println("Frame Size: " + frameSize);
```
fhcarter: change to log.debug or remove, please

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -240,31 +243,31 @@ public void testMetaConfig() {
         assert ypImageSaver.objectDetector.metaConfigOptions.useMetaIfAvailable == true;
         
         // Frame Size should be 416 since this is what is included in the meta file
```
fhcarter: comment should be 608, no?

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -948,18 +978,32 @@ public void testQuery() throws ImageProcessingException, InterruptedException {
...
+                    + "             \"metaFile\": \"" + META_FILE_608 + "\", "
+                    + "             \"type\": \"yolo\""
+                    + "         }"
+                    + "}";
 
-    List<Map> getExpectedResults() throws JsonParseException, JsonMappingException, IOException {
+    List<Map> getExpectedResults(boolean useDefault) throws JsonParseException, JsonMappingException, IOException {
         ObjectMapper m = new ObjectMapper();
-        return m.readValue(imageResultsAsString, List.class);
+        if (useDefault) {
```
fhcarter: This will be cleaner & more extensible if, instead of passing in a boolean, just pass in string to test.  That is,
```
List<Map> getExpectedResults(String expectedResults)
...
return m.readValue(expectedResults, List.class)
```

As soon as we have to test another image or another result, the boolean -- limit of two choices -- will be insufficient.

This can be done in a subsequent PR, but better to think a bit more future-proof.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -948,18 +978,32 @@ public void testQuery() throws ImageProcessingException, InterruptedException {
...
+            return m.readValue(imageResultsAsString, List.class);
+        } else {
+            return m.readValue(imageResultsAsString608, List.class);
+        }
     }
 
-    void verifyProcessing(YoloProcessor ypImageSaver) throws ImageProcessingException {
+    void verifyProcessing(YoloProcessor ypImageSaver, boolean useDefault) throws ImageProcessingException {
```
fhcarter: `String expectedResults` here, too

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -317,8 +328,8 @@ private Graph createNormalizerGraph() {
         	map.put("label", recognition.getTitle());
         	map.put("confidence", recognition.getConfidence());
         	
-        	float scaleX = (float) buffImage.getWidth() / (float) FRAME_SIZE;
-            float scaleY = (float) buffImage.getHeight() / (float) FRAME_SIZE;
+        	float scaleX = (float) buffImage.getWidth() / (float) frameSize;
+            float scaleY = (float) buffImage.getHeight() / (float) frameSize;
```
fhcarter: spacing/indent is off on 2nd line.  (Was before, too, but might as well fix)

objectRecognitionSource/src/main/java/edu/ml/tensorflow/classifier/YOLOClassifier.java
namirfa: My mistake, this was for my personal debugging. Removing these.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
namirfa: Correct, I'll change this.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
namirfa: Agreed this is a better model, I'll change it.

Issue #127, opened 2019-01-30
Object Recognition: Resizing Saved Images


namirfa:
Add the ability to configure the size of saved images. This configuration can be specified in the initial source configuration, and will also be configurable in the query parameters once Issue #126 is complete.

**NOTE:** We will only support *reducing* the dimensions of an image. We will *not* enlarge images.

Inspiration for this enhancement comes from network latency issues when trying to upload images to VANTIQ using only the saveImage and saveRate options. Adding the ability to lower the resolution of saved images when requested can create more flexibility for the user when network issues arise.

Issue #126, opened 2019-01-30
Object Recognition: Querying Enhancements


namirfa:
Add querying functionality for selecting locally saved images and uploading them to VANTIQ, or deleting them.

Inspiration for this enhancement comes from network latency issues when trying to upload images to VANTIQ using only the saveImage and saveRate options. Adding the ability to upload locally saved images when requested can create more flexibility for the user when network issues arise.

PR #128, opened 2019-01-31
Naf obj #127

namirfa:
Fixes Issue #127 

Adds the ability to specify "savedResolution" options in the source configuration. This field contains an option called "longEdge", which is used to resize the saved image. The longEdge must be a non-negative integer that is smaller than the long edge of the original image. The saved image will be resized according to the longEdge value, and maintain the same dimension ratio as the original image.

The tests utilize the vantiq.download feature to check the dimensions for both locally saved and VANTIQ saved images.

This feature has not been added to the query parameters yet. It will be added as part of the work required for Issue #126 .

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -161,7 +161,13 @@ The Configuration document may look similar to the following example:
              "metaFile": "coco-1.1.meta",
              "pbFile": "coco-1.1.pb",
              "type": "yolo",
-             "threshold": 0.2
+             "threshold": 0.2,
+             "saveImage": "both",
+             "saveRate": 1,
+             "outputDir": "imageOut",
+             "savedResolution": {
+                "longEdge": 400
+             }
           }
        }
     }
@@ -222,6 +228,11 @@ is set to be either "local" or "both".
 captured). If not specified, the value will default to 1 which saves every captured image.
 *   labelImage: Optional. If set to "true", images will be saved with bounding boxes and labels. If set to "false", or if not 
 set at all, the images will be saved with no bounding boxes or labels.
+*   savedResolution: Optional. A map containing options for adjusting the resolution of the saved images.
+    * *Options for savedResolution:*
+    1. longEdge: Optional. Used to resize the saved images. Must be a non-negative integer that is smaller than the long edge 
+    of the image to be saved. This value will become the new long edge dimension, and the short edge will be scaled down to 
+    maintain the same dimension ratio as the original image.
 
 ## Messages from the Source<a name="msgFormat" id="msgFormat"></a>
 
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -233,6 +233,8 @@ set at all, the images will be saved with no bounding boxes or labels.
     1. longEdge: Optional. Used to resize the saved images. Must be a non-negative integer that is smaller than the long edge 
     of the image to be saved. This value will become the new long edge dimension, and the short edge will be scaled down to 
     maintain the same dimension ratio as the original image.
+        *   **NOTE:** We do not support enlarging images. This feature can be used *only* to resize images to smaller 
+        dimensions.
 
 ## Messages from the Source<a name="msgFormat" id="msgFormat"></a>
 
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -28,6 +28,7 @@ public class ImageUtil {
     public Vantiq vantiq = null; // Added to allow image saving with VANTIQ
     public String sourceName = null;
     public Boolean saveImage;
+    public int longEdge = 0;
 
     /**
      * Label image with classes and predictions given by the ThensorFLow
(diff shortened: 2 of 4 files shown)


namirfa:
I think I can add more of the ImageUtil changes into one method, since the code checking the width vs. length etc. is redundant.

I'll work on that now.

objectRecognitionSource/README.md
```diff
@@ -222,6 +228,13 @@ is set to be either "local" or "both".
 captured). If not specified, the value will default to 1 which saves every captured image.
 *   labelImage: Optional. If set to "true", images will be saved with bounding boxes and labels. If set to "false", or if not 
 set at all, the images will be saved with no bounding boxes or labels.
+*   savedResolution: Optional. A map containing options for adjusting the resolution of the saved images.
+    * *Options for savedResolution:*
+    1. longEdge: Optional. Used to resize the saved images. Must be a non-negative integer that is smaller than the long edge
```
fhcarter: I think I'd change _Used to resize saved _images._ to _This sets the maximum long edge dimension for saved images._. By using this phraseology, we establish our _contract_ which then means that we don't have to worry about what happens if we save things that are smaller (because they were smaller), etc.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -63,7 +64,33 @@ public void saveImage(final BufferedImage image, final String target) {
...
             try {
                 IOUtil.createDirIfNotExists(new File(outputDir));
-                ImageIO.write(image,"jpg", new File(outputDir + File.separator + target));
+                if (longEdge == 0) {
+                    ImageIO.write(image, "jpg", new File(outputDir + File.separator + target));
+                } else {
+                    int width = image.getWidth();
+                    int height = image.getHeight();
+                    if (longEdge > Math.max(width, height)) {
+                        LOGGER.error("The longEdge value is too large, resizing cannot enlarge the original image. "
```
fhcarter: I think we should make this a trace() level note.  We've fulfilled our contract (see comment above on the readme) so there's nothing amiss here.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -63,7 +64,33 @@ public void saveImage(final BufferedImage image, final String target) {
...
+                                + "Original image resolution will not be changed");
+                        ImageIO.write(image, "jpg", new File(outputDir + File.separator + target));
+                        longEdge = 0;
```
fhcarter: This unsets longEdge for everything. Couldn't some source send back various images with different dimensions?  Seems like we just want to skip the resize, not change the expected behavior.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -89,7 +116,33 @@ public void saveImage(final BufferedImage image, final String target) {
...
+                                double ratio = (double) longEdge/height;
+                                int newWidth = (int) (width * ratio);
+                                BufferedImage resizedImage = resizeImage(newWidth, longEdge, image);
+                                ImageIO.write(resizedImage, "jpg", imgFile);
+                            } else {
+                                BufferedImage resizedImage = resizeImage(longEdge, longEdge, image);
+                                ImageIO.write(resizedImage, "jpg", imgFile);
+                            }
+                        }
+                    }
```
fhcarter: Seems like we should be able to refactor all this stuff in a method to call.  I think you're working on this.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -89,7 +116,33 @@ public void saveImage(final BufferedImage image, final String target) {
...
+                    if (longEdge == 0) {
+                        ImageIO.write(image, "jpg", imgFile);
+                    } else {
+                        int width = image.getWidth();
+                        int height = image.getHeight();
+                        if (longEdge > Math.max(width, height)) {
+                            LOGGER.error("The longEdge value is too large, resizing cannot enlarge the original image. "
+                                    + "Original image resolution will not be changed");
+                            ImageIO.write(image, "jpg", imgFile);
+                            longEdge = 0;
```
fhcarter: again, don't kill it forever

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -115,6 +168,21 @@ public void saveImage(final BufferedImage image, final String target) {
...
     }
+    
+    /**
+     * A function used to resize the original captured image, if requested.
+     * @param width     The resized width of the image to be saved.
+     * @param height    The resized height of the image to be saved.
+     * @param image     The original image to be resized.
+     * @return          The resized image.
+     */
+    public BufferedImage resizeImage(int width, int height, BufferedImage image) {
```
fhcarter: maybe we could have a variant that includes the longEdge that does all that work.

namirfa: Yep that's exactly what I'm doing now.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
namirfa: Yep, I'm working on this.

objectRecognitionSource/README.md
```diff
@@ -222,6 +228,13 @@ is set to be either "local" or "both".
 captured). If not specified, the value will default to 1 which saves every captured image.
 *   labelImage: Optional. If set to "true", images will be saved with bounding boxes and labels. If set to "false", or if not 
 set at all, the images will be saved with no bounding boxes or labels.
+*   savedResolution: Optional. A map containing options for adjusting the resolution of the saved images.
+    * *Options for savedResolution:*
+    1. longEdge: Optional. Used to resize the saved images. Must be a non-negative integer that is smaller than the long edge 
+    of the image to be saved. This value will become the new long edge dimension, and the short edge will be scaled down to 
+    maintain the same dimension ratio as the original image.
```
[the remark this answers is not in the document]
namirfa: I'll change that to aspect ratio. And we do have a test for the value being larger than the image. This is located in `TestYoloProcessor.testInvalidResizingSavedImage()`.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
namirfa: Just talked about this, forget the comment above :).

Issue #123, opened 2019-01-22
Object Recognition: Fix Query Tests for NoProcessor feature


namirfa:
Add tests that setup a source in VANTIQ, and then query against the source to demonstrate that querying is working as expected for the NoProcessor implementation.

namirfa:
These tests need to be set up for YOLO Processor as well.

PR #129, opened 2019-02-06
Naf query image #126

namirfa:
Fixes Issue #126 

Adds new functionality for Object Recognition Source queries. Queries can now be used to upload locally saved images to VANTIQ. Queries can also be used to delete locally saved images.

The images can be uploaded with user-specified resolution, as described in the README. This can be accomplished by adding a "savedResolution" option to the query parameters ("where" clause in query), exactly as you would for the source configuration.

**TO DO**:
* Need to update the README
* Need to handle the case of duplicate image names (i.e. 2019-02-05--02-35-10.jpg, 2019-02-05--02-35-10(1).jpg, 2019-02-05--02-35-10(2).jpg, etc...)

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
     instead of only the objects recognized. Note that the data will be the sole occupant of a 1-element array when
     received, instead of being immediately available as a JSON object. This is because Query results are mandated
     to be arrays. Default is false.
+    
+Unique query capabilities available for the YOLO Processor are:
+
+*   **NOTE:** All YOLO Processor queries must have an "operation" parameter specified. If this parameter is not specified, it 
+will be set to the default value, "processNextFrame".
+
+*   **Upload images to VANTIQ:**
+    *   The user can specify an image, or multiple images to be uploaded to VANTIQ as a document. The images will be those 
+    that are saved in the output directory, which is defined in the source configuration.
+    *   Parameters:
+        *   "operation": Required. Must be set to "upload".
+        *   *Only one of the following two values *MUST* be specified, or a query error will be returned. If both values are 
+        specified, the imageName value will be used. (For optimal use, only set one of these values.)*
+            *   "imageName": A string value representing the name of the file to be uploaded. If set to "all", then all images 
+            in the output directory will be uploaded.
+            *   "imageDate": A list containing two strings, a start date and an end date. All locally saved images falling 
+            between the start and end date, (inclusive), will be uploaded. Dates must be formatted in the following 
+            manner: "yyyy-MM-dd--HH-mm-ss".
+        *   "savedResolution": Optional. This value can be set in the exact same way as it is set in the source configuration. 
+        If it is defined here as a query parameter, it will override the value set in the source configuration.
+    
+*   **Delete locally saved images:**
+    *   The user can specify one, or multiple locally saved images to be deleted. The images will be those 
+    that are saved in the output directory, which is defined in the source configuration.
+    *   Parameters:
+        *   "operation": Required. Must be set to "delete".
+        *   *Only one of the following two values *MUST* be specified, or a query error will be returned. If both values are 
+        specified, the imageName value will be used. (For optimal use, only set one of these values.)*
+            *   "imageName": A string value representing the name of the file to be deleted. If set to "all", then all images 
+            in the output directory will be deleted.
+            *   "imageDate": A list containing two strings, a start date and an end date. All locally saved images falling 
+            between the start and end date, (inclusive), will be deleted. Dates must be formatted in the following 
+            manner: "yyyy-MM-dd--HH-mm-ss".
+
+*   **Process a single frame from the camera defined in the source configuration:**
+    *   Parameters:
+        *   "operation": Required. Must be set to "processNextFrame".
+        *   "NNsaveImage": Optional. This value can be set exactly like the "saveImage" value in the source configuration.
+        *   "NNoutputDir": Optional. This value can be set exactly like the "outputDir" value in the source configuration.
+        *   "NNfileName": Optional. A string representing the unique name used to save the file. If not specified, the file 
+        will be named using the standard <yyyy-MM-dd--HH-mm-ss.jpg> value.
+    
+**EXAMPLE QUERIES**:
+
+*   Upload Query using imageName:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"upload",
+    	imageName:"2019-02-08--10-33-36.jpg",
+    	savedResolution: {longEdge=600}
+```
+
+*   Upload Query using imageDate:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"upload",
+    	imageDate:["2019-02-08--10-33-36", "2019-02-08--10-34-06"]
+```
+
+*   Delete Query using imageName:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"delete",
+    	imageName:"all"
+```
+
+*   Delete Query using imageDate:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"delete",
+    	imageDate:["2019-02-08--10-33-36", "2019-02-08--10-34-06"]
+```
+
+*   Process Next Frame Query:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+        operation:"processNextFrame",
+    	NNsaveImage:"local",
+    	NNoutputDir:"testDir",
+    	NNfileName:"testFile"
+```
+
+*   Process Next Frame Query without specifying operation *(not recommended)*:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	NNsaveImage:"local",
+    	NNoutputDir:"testDir",
+    	NNfileName:"testFile"
+```
 
 ### Error Messages
 
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -353,7 +353,7 @@ SELECT * FROM SOURCE Camera1 AS results WITH
     	imageName:"2019-02-08--10-33-36.jpg"
 ```
 
-*   Delete Query using imageDate to save everything after a certain date:
+*   Delete Query using imageDate to delete everything after a certain date:
 ```
 SELECT * FROM SOURCE Camera1 AS results WITH
     	operation:"delete",
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -286,8 +286,8 @@ Unique query capabilities available for the YOLO Processor are as follows:
 it will be set to the default value, "processNextFrame".
 
 *   **Upload images to VANTIQ:**
-    *   The user can specify an image, or multiple images to be uploaded to VANTIQ as a document. The images will be those 
-    that are saved in the output directory, which is defined in the source configuration.
+    *   The user can specify an image or a set of images specified by their date & time to be uploaded to VANTIQ as a 
+    document. The images will be those that are saved in the output directory, which is defined in the source configuration.
     *   Parameters:
         *   "operation": Required. Must be set to "upload".
         *   *You should specify exactly one of the following two values. If neither is specified, a query error will be 
@@ -299,14 +299,16 @@ it will be set to the default value, "processNextFrame".
                 *   To select all images *before* or *after* a certain date, the "-" value can be used as one of the date 
                 strings in the list. For example, the following would save all dates *before* the given date: 
                     *   \["-", yourEndDate\]. 
-                *   To select *all files* in the output directory, one could use the following value for imageDate: 
+                *   To select *all files* in the output directory, one must use the following value for imageDate: 
                     *   \["-", "-"\].
-        *   "savedResolution": Optional. This value can be set in the exact same way as it is set in the source configuration. 
-        If it is defined here as a query parameter, it will override the value set in the source configuration.
+        *   "savedResolution": Optional. This value can be set in the same way as it is set in the source configuration. If it 
+        is defined here as a query parameter, it will override the value set in the source configuration, otherwise the source 
+        configuration value will be used. The setting cannot be larger than that provided in the source configuration (since 
+        that defines how the images are saved).
     
 *   **Delete locally saved images:**
-    *   The user can specify one, or multiple locally saved images to be deleted. The images will be those 
-    that are saved in the output directory, which is defined in the source configuration.
+    *   The user can specify an image or a set of images specified by their date & time to be deleted. The images will be 
+    those that are saved in the output directory, which is defined in the source configuration.
     *   Parameters:
         *   "operation": Required. Must be set to "delete".
         *   *You should specify exactly one of the following two values. If neither is specified, a query error will be 
@@ -318,7 +320,7 @@ it will be set to the default value, "processNextFrame".
                 *   To select all images *before* or *after* a certain date, the "-" value can be used as one of the date 
                 strings in the list. For example, the following would save all dates *before* the given date: 
                     *   \["-", yourEndDate\]. 
-                *   To select *all files* in the output directory, one could use the following value for imageDate: 
+                *   To select *all files* in the output directory, one must use the following value for imageDate: 
                     *   \["-", "-"\].
 
 *   **Process a single frame from the camera defined in the source configuration:**
@@ -358,6 +360,13 @@ SELECT * FROM SOURCE Camera1 AS results WITH
 SELECT * FROM SOURCE Camera1 AS results WITH
     	operation:"delete",
     	imageDate:["2019-02-08--10-33-36", "-"]
+```
+
+*   Delete Query using imageDate to delete all images:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"delete",
+    	imageDate:["-", "-"]
 ```
 
 *   Process Next Frame Query:
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -28,6 +28,7 @@ public class ImageUtil {
     public Vantiq vantiq = null; // Added to allow image saving with VANTIQ
     public String sourceName = null;
     public Boolean saveImage;
+    public int frameSize;
     public int longEdge = 0;
 
     /**
@@ -36,8 +37,8 @@ public class ImageUtil {
      * @param recognitions  list of recognized objects
      */
     public BufferedImage labelImage(BufferedImage bufferedImage, final List<Recognition> recognitions) {
-        float scaleX = (float) bufferedImage.getWidth() / (float) Config.FRAME_SIZE;
-        float scaleY = (float) bufferedImage.getHeight() / (float) Config.FRAME_SIZE;
+        float scaleX = (float) bufferedImage.getWidth() / (float) frameSize;
+        float scaleY = (float) bufferedImage.getHeight() / (float) frameSize;
         Graphics2D graphics = (Graphics2D) bufferedImage.getGraphics();
 
         for (Recognition recognition: recognitions) {
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -29,6 +29,7 @@ public class ImageUtil {
     public String sourceName = null;
     public Boolean saveImage;
     public int frameSize;
+    public Boolean queryResize = false;
     public int longEdge = 0;
 
     /**
@@ -59,6 +60,7 @@ public class ImageUtil {
      * and erroring out if it disappears in the interim 
      * @param image     The image to save
      * @param target    The name of the file to be written
+     * @throws IOException 
      */
     public void saveImage(final BufferedImage image, final String target) {
         File fileToUpload = null;
@@ -78,20 +80,7 @@ public class ImageUtil {
         }
         if (vantiq != null) {
             if (fileToUpload != null) {
-                vantiq.upload(fileToUpload, 
-                        "image/jpeg", 
-                        "objectRecognition/" + sourceName + '/' + target,
-                        new BaseResponseHandler() {
-                            @Override public void onSuccess(Object body, Response response) {
-                                super.onSuccess(body, response);
-                                LOGGER.trace("Content Location = " + this.getBodyAsJsonObject().get("content"));
-                            }
-                            
-                            @Override public void onError(List<VantiqError> errors, Response response) {
-                                super.onError(errors, response);
-                                LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
-                            }
-                });
+                uploadImage(fileToUpload, target);
             } else {
                 try {
                     File imgFile = File.createTempFile("tmp", ".jpg");
@@ -102,29 +91,79 @@ public class ImageUtil {
                         BufferedImage resizedImage = resizeImage(image);
                         ImageIO.write(resizedImage, "jpg", imgFile);
                     }
-                    vantiq.upload(imgFile, 
-                            "image/jpeg", 
-                            "objectRecognition/" + sourceName + '/'  + target,
-                            new BaseResponseHandler() {
-                                @Override public void onSuccess(Object body, Response response) {
-                                    super.onSuccess(body, response);
-                                    LOGGER.trace("Content Location = " + this.getBodyAsJsonObject().get("content"));
-                                    if(imgFile.delete()) { 
-                                        LOGGER.trace("Temp file deleted successfully"); 
-                                    } else { 
-                                        LOGGER.warn("Failed to delete temp file"); 
-                                    } 
-                                }
-                                
-                                @Override public void onError(List<VantiqError> errors, Response response) {
-                                    super.onError(errors, response);
-                                    LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
-                                }
-                    });
+                    uploadImage(imgFile, target);
                 } catch (IOException e) {
                     LOGGER.error("Unable to save image {}", target, e);
                 }
             }
+        }
+    }
+    
+    /**
+     * A method used to upload images to VANTIQ, using the VANTIQ SDK
+     * @param imgFile   The file to be uploaded. If not specified to save locally, this file will be deleted.
+     * @param target    The name of the file to be uploaded.
+     * @throws IOException 
+     */
+    public void uploadImage(File imgFile, String target) {
+        if (queryResize) {
+            try {
+                BufferedImage resizedImage = resizeImage(ImageIO.read(imgFile));
+                File tmpFile = File.createTempFile("tmp", ".jpg");
+                tmpFile.deleteOnExit();
+                ImageIO.write(resizedImage, "jpg", tmpFile);
+                vantiq.upload(tmpFile, 
+                        "image/jpeg", 
+                        "objectRecognition/" + sourceName + '/'  + target,
+                        new BaseResponseHandler() {
+                            @Override public void onSuccess(Object body, Response response) {
+                                super.onSuccess(body, response);
+                                LOGGER.trace("Content Location = " + this.getBodyAsJsonObject().get("content"));
+                                
+                                if (outputDir == null) {
+                                    deleteImage(imgFile);
+                                }
+                            }
+                            
+                            @Override public void onError(List<VantiqError> errors, Response response) {
+                                super.onError(errors, response);
+                                LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
+                            }
+                });
+            } catch (IOException e) {
+                LOGGER.error("An error occured while reading the locally saved image file. " + e.getMessage());
+            }
+        } else {
+            vantiq.upload(imgFile, 
+                    "image/jpeg", 
+                    "objectRecognition/" + sourceName + '/'  + target,
+                    new BaseResponseHandler() {
+                        @Override public void onSuccess(Object body, Response response) {
+                            super.onSuccess(body, response);
+                            LOGGER.trace("Content Location = " + this.getBodyAsJsonObject().get("content"));
+                            
+                            if (outputDir == null) {
+                                deleteImage(imgFile);
+                            }
+                        }
+                        
+                        @Override public void onError(List<VantiqError> errors, Response response) {
+                            super.onError(errors, response);
+                            LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
+                        }
+            });
+        }
+    }
+    
+    /**
+     * A method used to delete locally saved images.
+     * @param imgFile  The file to be deleted.
+     */
+    public void deleteImage(File imgFile) {
+        if(imgFile.delete()) {
+            LOGGER.trace("File was successfully deleted.");
+        } else {
+            LOGGER.error("Failed to delete file");
         }
     }
     
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -1,6 +1,5 @@
 package edu.ml.tensorflow.util;
 
-import edu.ml.tensorflow.Config;
 import edu.ml.tensorflow.model.BoxPosition;
 import edu.ml.tensorflow.model.Recognition;
 import io.vantiq.client.BaseResponseHandler;
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
@@ -104,53 +104,45 @@ public class ImageUtil {
      * @param target    The name of the file to be uploaded.
      */
     public void uploadImage(File imgFile, String target) {
+        File fileToUpload = imgFile;
         if (queryResize) {
             try {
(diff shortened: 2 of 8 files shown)


objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -92,10 +92,27 @@ public void handleMessage(ExtensionServiceMessage message) {
...
-                    source.sendDataFromImage(data, message);
+                Map<String, ?> request = (Map<String, ?>) message.getObject();
+                String replyAddress = ExtensionServiceMessage.extractReplyAddress(message);
+                if (request.get("operation") instanceof String) {
+                    String operation = (String) request.get("operation");
+                    if (operation.equalsIgnoreCase("upload")) {
+                        source.uploadLocalImages(request, replyAddress);
+                    } else if (operation.equalsIgnoreCase("delete")) {
+                        source.deleteLocalImages(request, replyAddress);
+                    } else {
```
fhcarter: Suppose I send an operation of "nutzo".  I think it will default to something here instead of returning in invalid operation -- which, I think, it should do.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -92,10 +92,27 @@ public void handleMessage(ExtensionServiceMessage message) {
...
+                        // Read, process, and send the image
+                        ImageRetrieverResults data = source.retrieveImage(message);
+                        if (data != null) {
+                            source.sendDataFromImage(data, message);
+                        }
```
fhcarter: again, avoid duplication.

```
// get the operation
// if none specified, set operation to the default operation code
//, then do the work based on the operation.

File: objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java [hunk 3]
```diff
@@ -101,32 +89,81 @@ public void saveImage(final BufferedImage image, final String target) {
...
+                            }
+                            
+                            @Override public void onError(List<VantiqError> errors, Response response) {
+                                super.onError(errors, response);
+                                LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
+                            }
+                });
+            } catch (IOException e) {
+                LOGGER.error("An error occured while reading the locally saved image file. " + e.getMessage());
+            }
```
fhcarter: Seems like most of this code is duplicated.  Couldn't this do
```
fileToUpload = imgFile
if (queryResize) {
    ... do the resize...
    fileToUpload = tmpFile
} 
// Now do the upload of fileToUpload
```

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,362 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+    * @param imageUtil   The instantiated ImageUtil class containing the method to upload
+    * @param filter      The filter used if a dateRange was selected, otherwise null
+    */
+   public void uploadMany(String imageDir, ImageUtil imageUtil, FilenameFilter filter) {
+       File imgDirectory = new File(imageDir);
+       File[] directoryListing;
+       if (filter == null) {
+           directoryListing = imgDirectory.listFiles();
+       } else {
+           directoryListing = imgDirectory.listFiles(filter);
```
fhcarter: Could this be simplified by setting up a filter that selects everything?

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestUploadAndDeleteQuery.java
```diff
@@ -0,0 +1,612 @@
...
+    public void testInvalidImageUploadParameters() throws InterruptedException {
+        // Only run test with intended vantiq availability
+        assumeTrue(testAuthToken != null && testVantiqServer != null);
+        
+        // Forgetting to include imageDir
+        Map<String, Object> request = new LinkedHashMap<>();
+        request.put("imageName", "all");
+        core.uploadLocalImages(request, null);
+                
+        // Checking that all images were uploaded to VANTIQ
```
fhcarter: "all the images are still there?" or aren't there?  Didn't these get inserted in setup()

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,362 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
        }
    }
    
+   /**
+    * Uploads requested images to VANTIQ. Called when query parameter "operation" is set to "save".
+    * @param request        The parameters sent with the query.
+    * @param replyAddress   The replyAddress used to send a query response.
+    */
+   public void uploadLocalImages(Map<String, ?> request, String replyAddress) {
```
fhcarter: Here, there's a certain amount of duplicated code plus a bunch of mixed processing.  Seems like there might be a better ways to structure things.  Something like:

// Examine parameters
// Check completeness/legality/etc.
// Compute a filter for appropriate parameter set
// Make call to upload_Whatever_ based on filter (and, possibly, if single)

Issue is that right now, there's a very long routine that's doing a whole bunch of different things intermixed.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,362 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+           }
+       }
+   }
+   
+   /**
+    * Deletes requested locally-saved images. Called when query parameter "operation" is set to "delete".
+    * @param request        The parameters sent with the query.
+    * @param replyAddress   The replyAddress used to send a query response.
+    */
+   public void deleteLocalImages(Map<String, ?> request, String replyAddress) {
```
fhcarter: similar comment to uploadLocalImages.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestUploadAndDeleteQuery.java
```diff
@@ -0,0 +1,612 @@
...
+        
+        Map<String, Object> request = new LinkedHashMap<>();
+        request.put("imageDir", OUTPUT_DIR);
+        request.put("imageName", START_DATE);
+        
+        core.uploadLocalImages(request, null);
+                
+        // Checking that all images were uploaded to VANTIQ
+        Thread.sleep(1000);
+        checkUploadToVantiq(IMAGE_NAME_2);
```
fhcarter: Up to you here, but I find it useful to create a map that has all the related stuff together.  Thus, IMAGE_2 might be a map that has a file name, a date specification, etc., all in the same place.  Easier to track if/when things change later...
Optional, as I said.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestUploadAndDeleteQuery.java
```diff
@@ -0,0 +1,612 @@
...
+        }
+    }
+    
+    public void checkNotUploadToVantiq(String name) {
+        vantiqResponse = vantiq.selectOne("system.documents", name);
+        if (vantiqResponse.isSuccess()) {
+            fail("Image should not have been uploaded to VANTIQ");
+        }
+    }
+}
```
fhcarter: I may have missed it, but I don't see any tests of the actual queries, operations, invalid things, etc.  Seems like we could use the SDK to define procedures & execute them to verify that we got what we expect (or errors when we don't expect).  As these things get more & more complicated, better to have a fuller test system.

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
namirfa: The reason I implemented this with redundant code is because I was getting the following error on line 121 (deleteImage(imgFile) - "Local variable statement defined in an enclosing scope must be final or effectively final."

I can't re-assign the value of imgFile because then the onSuccess callback function cannot delete the temporary file if it needs to. Let me know if that makes sense!

fhcarter: Yeah, don't try & reassign the parameter -- confusing anyway, if call by value vs name vs reference, etc.

IntelliJ seems happy with
```
   public void uploadImage(File imgFile, String target) {
        File fileToUpload = imgFile;
        if (queryResize) {
            try {
                BufferedImage resizedImage = resizeImage(ImageIO.read(imgFile));
                File tmpFile = File.createTempFile("tmp", ".jpg");
                tmpFile.deleteOnExit();
                ImageIO.write(resizedImage, "jpg", tmpFile);
                fileToUpload = tmpFile;

            } catch (IOException e) {
                LOGGER.error("An error occured while reading the locally saved image file. " + e.getMessage());
            }
        } 
        vantiq.upload(fileToUpload,
                "image/jpeg",
                "objectRecognition/" + sourceName + '/'  + target,
                new BaseResponseHandler() {
                    @Override public void onSuccess(Object body, Response response) {
                        super.onSuccess(body, response);
                        LOGGER.trace("Content Location = " + this.getBodyAsJsonObject().get("content"));

                        if (outputDir == null) {
                            deleteImage(imgFile);
                        }
                    }

                    @Override public void onError(List<VantiqError> errors, Response response) {
                        super.onError(errors, response);
                        LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
                    }
                });
    }
```

objectRecognitionSource/src/main/java/edu/ml/tensorflow/util/ImageUtil.java
```diff
@@ -101,32 +89,81 @@ public void saveImage(final BufferedImage image, final String target) {
...
+                                }
+                            }
+                            
+                            @Override public void onError(List<VantiqError> errors, Response response) {
+                                super.onError(errors, response);
+                                LOGGER.error("Errors uploading image with VANTIQ SDK: " + errors);
+                            }
+                });
+            } catch (IOException e) {
+                LOGGER.error("An error occured while reading the locally saved image file. " + e.getMessage());
```
fhcarter: Error's not necessarily correct.  Might've been with resizing, etc.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestUploadAndDeleteQuery.java
namirfa: Almost done with fixes for all the other comments. For this addition to the tests, I was going to add it as part of the fixes for Issue #123 . Let me know if it's alright to add those tests all together, or if I should add them now.

namirfa:
Hey @fhcarter 

I took a stab at refactoring the objectRecognitionCore code, but in all honesty I am not sure if I ended up making things even more confusing. If you have time to quickly review that and let me know how I can improve it, that would be much appreciated.

I didn't change the deleteLocalImages method yet, since I'm not 100% confident in the changes I have made to uploadLocalImages().

I added fixes for all of your other comments, and still need to address the README and * duplicate image names I addressed in the initial comment.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,408 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+    */
+   public void uploadLocalImages(Map<String, ?> request, String replyAddress) {   
+       String imageDir = null;
+       String imageName = null;
+       String imageDate = null;
+       String dateRange = null;
+       List<String> imageDateList = null;
+       ImageUtil imageUtil;
+       
+       // ------------------------ GATHERING PARATMERS ---------------------------
```
fhcarter: Parameters :-)

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,408 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+       
+       // Checking to make sure there is an image directory specified. (Required)
+       if (request.get("imageDir") instanceof String) {
+           imageDir = (String) request.get("imageDir");
+           
+           // Set up ImageUtil properties
+           imageUtil = setupQueryImageUtil(request);
```
fhcarter: Personally, I would probably structure this as "detect negatives then deal with it".  So something like
```
if (!imageDir) {
  send error
} else {
   imageDir = ...
}
imageDate = ...
dateRange = ...
imageName = ...

if (imageName) {
  thingsByName = true
} else {
  if (imageDate & dateRange don't work together) {
    sendError
}
// Now we know things are kosher, deal with them....
```

One thing that might be worth considering:  What if we said that imageDate was always a list.  Then, we short circuit a bunch of this.  If it's size is 2, then we know what's up.  Otherwise, we may have to guess.

that said, what if imageDate was always a list of strings.  Usually, these strings are dates.  But, we also allow "none" or "-" or something to indicate that there's no date.  Then, we turn it into a list of Dates (or date strings or whatever) where the "-" means no date (== -infinity or +infinity).  Then we just use that specification to figure out what we're doing.

Something like that may be less complex to explain -- don't know if it would necessarily be easier to implement, but it probably is because we just check for our list & look for special values.  Things other than "-" (which is probably more international than "none") and valid dates would be right out.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,408 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+               imageDateList = (List<String>) request.get("imageDate");
+               if (imageDateList.size() != 2) {
+                   client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                           "The imageDate value was a list with more than two elements. Must be a list containing only "
+                           + "[<yourStartDate>, <yourEndDate>].", null);
+                   return;
+               }
+           } else {
+               client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                       "No imageName or imageDate was specified, or they were incorrectly specified. "
```
fhcarter: Another option to dump all these else's with code in them is to have the various error checkers just set set a code & message (& params if required).  Then, one place at the bottom (or wherever we're sure there's no other error), we make a single call to `client.sendQueryError(replyAddr, code, errorText, errorParams)`. Sometimes that makes things a bit clearer.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,408 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+           } else {
+               uploadOne = true;
+               name = imageName;
+           }
+       } else if (imageDate != null) {
+           if (imageDate.equals("all")) {
+               uploadOne = false;
+           } else if(dateRange.equals("none")) {
+               uploadOne = true;
+               name = imageDate;
```
fhcarter: Here, if our imageDate is always a list, we create a single 
`createFilter(imageDate)` that just constructs the filter itself.

Or just create a new class
```
class DateRangeFilter implements FileNameFilter {
  beforeDate = null
  afterDate = null
   public DateRangeFilter(List<Date> dateRange) {
      // We assume this list has already been checked for size & dates converted
      beforeDate = dateRange[0] // Either (or both) of these could be null...
      afterDate = dateRange[1]
  }

  @Override
  public boolean accept(File dir, String name) {
        fName =format.parse(name)
			return ((beforeDate == null || fName.after(beforeDate)) &&
                             (afterDate == null || fName.before(afterDate))
		}
    }
}
```
Or something like that (not sure if the logic is quite right, but you get the gist.  The notion being that we'll take the imageDate list & convert it to a list of (2) Dates, either of which could be null.  If both are null, then it's the universal filter.  Otherwise, it does what we need.

The code will look very simple as compared to now, but we'll need a nice explanation of how the filter actually works.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +399,408 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+   
+   /**
+    * A helper function called by processImageQueryParameters, used to upload one specific image
+    * @param name       The name of the image to be uploaded
+    * @param imageDir   The name of the image directory
+    * @param imageUtil  The instantiated ImageUtil class containing the method to upload
+    */
+   public void uploadOne(String name, String imageDir, ImageUtil imageUtil) {
+       if (!name.endsWith(".jpg")) {
+           name = name + ".jpg";
```
fhcarter: Do we need to deal with non JPEG images?  What if we get a PNG thing in?

namirfa:
@fhcarter 

I refactored both the uploadLocalImages() and deleteLocalImages() methods based on your advice. I think it looks much cleaner now, and makes the "imageDate" parameter a bit more straightforward.

Let me know how it looks! If it looks alright, I'll add the check for duplicate image names, and I'll update the README to reflect these changes.

namirfa:
@fhcarter 

I added fixes for all of your review comments. Last things to do will still be updating the README and handling duplicate filenames.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +401,302 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+       // Checking to make sure there is an image directory specified. (Required)
+       if (!(request.get("imageDir") instanceof String)) {
+           client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                   "No imageDir was specified, or imageDir was incorrectly specified. "
+                   + "Cannot select image(s) to be "+ operation + " .", null);
+           return null;
+       } else {
+           imageDir = (String) request.get("imageDir");
+           // Set up ImageUtil properties
+           imageUtil = setupQueryImageUtil(request);
```
fhcarter: This is not where I'd expect this to happen, but that's the way it works.  We can add this to the list of restructuring necessary.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +401,302 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+       } else if (imageDate != null) {
+           for (String date : imageDate) {
+               if (date.equals("-")) {
+                   dateRange.add(null);
+               } else {
+                   try {
+                       dateRange.add(format.parse(date));
+                   } catch (ParseException e) {
+                       client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                               "One of the dates in the imageDate list could not be parsed.Please be sure that both "
```
fhcarter: need space in "parsed.Please"

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +401,302 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+                       dateRange.add(format.parse(date));
+                   } catch (ParseException e) {
+                       log.error("An error occurred while parsing the imageDate");
+                       client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                               "One of the dates in the imageDate list could not be parsed.Please be sure that both "
+                               + "dates are in the following format: yyyy-MM-dd--HH-mm-ss", null);
+                       return;
+                   }
+               }
+           }
```
fhcarter: Is there some reason we cannot do all this processing in the handleQueryParameters() call?  That should return, in its map, either an imageName set OR an imageDate that would be a filter.  Then, if imageName is set, we upload/deleteOne (which may no longer be necessary);  otherwise, use the filter.  We may have to send in an "intent" if we need to parameterize the errors to say for "update" vs. "delete".

We have literally the same copied code both places (except for uploadOne vs deleteOne -- both which, if necessary, become "isSingle" or something).  text error compatible.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +401,302 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+           uploadOne = false;
+       }
+       
+       // Calling helper functions to upload images (using filters if needed)
+       
+       if (uploadOne) {
```
fhcarter: I think `uploadOne` goes away -- this becomes `if (imageName != null)`...

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -390,6 +401,302 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+                   } catch (ParseException e) {
+                       client.sendQueryError(replyAddress, "io.vantiq.extsrc.objectRecognition.invalidQueryRequest", 
+                               "One of the dates in the imageDate list could not be parsed.Please be sure that both "
+                               + "dates are in the following format: yyyy-MM-dd--HH-mm-ss", null);
+                       return;
+                   }
+               }
+           }
+           deleteOne = false;
+       }
```
fhcarter: more cloned code -- move into `handleQueryParameters()`

namirfa:
@fhcarter 

I have updated the code based on your reviewer comments. I also added a check in the DateNameFilter to handle any duplicate filenames. I am editing the README now.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -212,6 +234,9 @@ boolean createNeuralNet(Map<String, ?> neuralNetConfig ) {
             log.debug("No neural net type specified. Trying for default of '{}'", DEFAULT_NEURAL_NET);
         }
         
+        // Setting the outputDir value for the core, (null if it does not exist)
+        source.outputDir = (String) neuralNetConfig.get(NeuralNetInterface.OUTPUT_DIRECTORY_ENTRY);
```
fhcarter: Minor, but I thought we were going to add the "type" to the interface as well, and use it a few lines above this.

namirfa:
@fhcarter 

I have edited the README to reflect the new changes.

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
     instead of only the objects recognized. Note that the data will be the sole occupant of a 1-element array when
     received, instead of being immediately available as a JSON object. This is because Query results are mandated
     to be arrays. Default is false.
+    
+Unique query capabilities available for the YOLO Processor are:
```
fhcarter: ... are as follows:

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+
+*   **NOTE:** All YOLO Processor queries must have an "operation" parameter specified. If this parameter is not specified, it
```
fhcarter: s/must/should/ -- since we say 1) it must be there, and 2) if it's not, we...

So it's a _should_ with a default.

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+
+*   **NOTE:** All YOLO Processor queries must have an "operation" parameter specified. If this parameter is not specified, it 
+will be set to the default value, "processNextFrame".
+
+*   **Upload images to VANTIQ:**
+    *   The user can specify an image, or multiple images to be uploaded to VANTIQ as a document. The images will be those 
+    that are saved in the output directory, which is defined in the source configuration.
+    *   Parameters:
+        *   "operation": Required. Must be set to "upload".
+        *   *Only one of the following two values *MUST* be specified, or a query error will be returned. If both values are
```
fhcarter: This section is inconsistent:  first, it says "only one" otherwise an error. Then it says, if both, we'll pick one.  We need to choose between these behaviors.

fhcarter: Perhaps it's trying to say that _at least one_ must be there.  But we're really saying that _exactly one_ should be there.  But if you say both, we'll pick the more specific?

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+    *   The user can specify an image, or multiple images to be uploaded to VANTIQ as a document. The images will be those 
+    that are saved in the output directory, which is defined in the source configuration.
+    *   Parameters:
+        *   "operation": Required. Must be set to "upload".
+        *   *Only one of the following two values *MUST* be specified, or a query error will be returned. If both values are 
+        specified, the imageName value will be used. (For optimal use, only set one of these values.)*
+            *   "imageName": A string value representing the name of the file to be uploaded. If set to "all", then all images 
+            in the output directory will be uploaded.
+            *   "imageDate": A list containing two strings, a start date and an end date. All locally saved images falling 
+            between the start and end date, (inclusive), will be uploaded. Dates must be formatted in the following
```
fhcarter: need to define the "-" date.

Also, would it be better to do "all" as` imageDate: [-,-]`? rather than a special name (that's what we really do anyway :-)

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+    	NNfileName:"testFile"
+```
+
+*   Process Next Frame Query without specifying operation *(not recommended)*:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	NNsaveImage:"local",
+    	NNoutputDir:"testDir",
+    	NNfileName:"testFile"
+```
```
fhcarter: Need at least one example with open end. Like `imageDate: ["-", "2019-02-08--10-33-36"]`

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/NeuralNetInterface.java
```diff
@@ -18,6 +18,7 @@
  * An interface for the neural net that will process the image and return a List of data representing the objects found.
  */
 public interface NeuralNetInterface {
+    static final String OUTPUT_DIRECTORY_ENTRY = "outputDir";
```
fhcarter: need to add (& use) "type" here.

objectRecognitionSource/README.md
fhcarter: Note that according to the tests, that's how it's done.  We don't seem to have a test for "all" as the imageName.  So maybe we should just remove it.  Use the range of _beginning of time_ to _end of time_ instead.

If we _do not_ make this change, we need to add test for `imageName: "all"`

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"upload",
+    	imageDate:["2019-02-08--10-33-36", "2019-02-08--10-34-06"]
+```
+
+*   Delete Query using imageName:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"delete",
+    	imageName:"all"
```
fhcarter: note -- we have no tests for this.

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,95 @@ Options available for all Queries (not prepended by anything) are:
...
+        will be named using the standard <yyyy-MM-dd--HH-mm-ss.jpg> value.
+    
+**EXAMPLE QUERIES**:
+
+*   Upload Query using imageName:
+```
+SELECT * FROM SOURCE Camera1 AS results WITH
+    	operation:"upload",
+    	imageName:"2019-02-08--10-33-36.jpg",
+    	savedResolution: {longEdge=600}
```
fhcarter: Does `{longEdge=600}` work?  Doesn't it have to be `{longEdge: 600}`

We need to have tests (maybe part of another test that we need to get in place soon) that test all the examples.  We should include these, or some variation of them, in the sample project.

Working code wins, so we should verify all these queries (albeit with different dates/times)

objectRecognitionSource/README.md
namirfa: I'm addressing all of your other comments now. We actually do have a test that checks imageName: "all".  It is called `testImageNameUploadAll()`. In any case, I will definitely add the "-" date.

I can remove the "all" behavior entirely if that will simplify things, let me know if you think that's worth doing.

objectRecognitionSource/README.md
namirfa: I think I will change this to "at least one". Right now, if you specify both then it will pick the more specific one, which is imageName.

objectRecognitionSource/README.md
namirfa: I actually ran this query and had longEdge set with an "=", (which was a mistake), but it did work as expected.

Must be why I didn't notice the swap of "=" for ":".

I will change this in the example to be a colon, but it is interesting that this query works with an equals sign as well.

objectRecognitionSource/README.md
fhcarter: We should probably say that we want exactly one.  But say it like "you should specify exactly one of the following..."  then we can give the behavior if you screw up.

objectRecognitionSource/README.md
fhcarter: yes, I missed that test.  Sorry.

But I think it might be simpler & clearer to just use the date range of negative-positive infinity instead of "all"

objectRecognitionSource/README.md
fhcarter: compiling it down to groovy forgives a multitude of sins :-)

namirfa:
@fhcarter 

Added fixes for all your reviewer comments. I did not add the example queries to the sample project yet. There is another issue filed which will add more thorough query tests (using the VANTIQ SDK) for all YOLO queries and No Processor queries. I will add the example queries to the sample project as part of that issue.

fhcarter:
Note that this PR is listed as having conflict with ImageUtil.java -- need to merge this in & resolve before (hopefully final) review.

namirfa:
@fhcarter 

Resolving that conflict should have been easy, but it all got confused. In any case, it is all resolved now and everything is in place. 

The project builds successfully, and all the tests pass. The README changes are in place as well.

objectRecognitionSource/README.md
fhcarter: I think we should add one that shows how to delete or upload everything.

*   Delete Query using imageDate to delete all images:
	```
	SELECT * FROM SOURCE Camera1 AS results WITH
	    	operation:"delete",
	    	imageDate:["-", "-"]
	```

objectRecognitionSource/README.md
```diff
@@ -279,6 +279,103 @@ Options available for all Queries (not prepended by anything) are:
...
+            *   "imageDate": A list containing two strings, a start date and an end date. All locally saved images falling 
+            between the start and end date, (inclusive), will be uploaded. Dates must be formatted in the following 
+            manner: "yyyy-MM-dd--HH-mm-ss".
+                *   To select all images *before* or *after* a certain date, the "-" value can be used as one of the date 
+                strings in the list. For example, the following would save all dates *before* the given date: 
+                    *   \["-", yourEndDate\]. 
+                *   To select *all files* in the output directory, one could use the following value for imageDate: 
+                    *   \["-", "-"\].
+        *   "savedResolution": Optional. This value can be set in the exact same way as it is set in the source configuration. 
+        If it is defined here as a query parameter, it will override the value set in the source configuration.
```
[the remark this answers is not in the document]
namirfa: One thing I was thinking of adding to the README, but not sure if it's worth writing out. Say someone set their source configuration to save images locally, with a longEdge of 600. The images would already be resized, and so you would only be able to set a longEdge less than the initial 600 from the source configuration.

Does that make sense? I'm wondering if I should put this in writing somewhere in the README, or if it is already implied since we said earlier that we will not support enlarging images.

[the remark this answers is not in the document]
fhcarter: I think it makes sense to add.  I think it's hard to explain all the ins & outs, so probably something like "the setting cannot be larger than that provided in the source configuration (since that defines how the images are saved)."  That gets to what's real without having to explain the various machinations that may go on...

PR #131, opened 2019-02-07
Naf label fix #130

### LINKED ISSUE #130: Object Recognition: Incorrect Labelling of Bounding Boxes
The bounding box values are calculated correctly, and sent correctly as part of the of the returned JSON.

However, the boxes are not labelled on the image correctly. This resulted from the enhancement made earlier, which allowed for the input Frame Size to be configurable. The labelling still uses the old frame size value, not the new one.

Once this has been fixed, the label on the image should appear correctly.

namirfa:
Fixes Issue #130 

Changes the labelImage() method in the ImageUtil class to use the correct frameSize value. Added fix for both polling and queries (processNextFrame). 

All tests pass, and the project builds successfully. I will add the configurable bounding box size as a separate enhancement issue.

### DIFF
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -216,6 +216,7 @@ public class ObjectDetector {
                 imageUtil.outputDir = outputDir;
                 imageUtil.vantiq = vantiq;
                 imageUtil.sourceName = sourceName;
+                imageUtil.frameSize = frameSize;
                 if (fileName == null) {
                     fileName = format.format(now) + ".jpg";
                 } else if (!fileName.endsWith(".jpg") && !fileName.endsWith(".jpeg")) {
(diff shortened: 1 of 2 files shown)


objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -127,6 +127,7 @@ public ObjectDetector(float thresh, String graphFile, String labelFile, String m
                 }
             }
             this.imageUtil = imageUtil;
+            imageUtil.frameSize = this.frameSize;
```
fhcarter: I think this is Ok to get this done, but I'm confused (in reading it all) as to who's in charge of what.  We have, in ODCore, the setupImageUtil() that constructs the imageUtil, but, apparently, gets the frame size wrong.  So we go in after the fact and stuff in a new value.  Seems like that should be either a parameter that it gets at startup OR it should know to go elsewhere (maybe it's metaconfig) to find it.  

The problem with "lots of folks" poking various public things is that it's very hard to track down what's going, where, etc.  I realize we're attempting to deal with some inherited code here, having it deal with things it wasn't built to do, but it'll prove problematic in the long run.

PR #135, opened 2019-02-13
Adding tests that use VANTIQ SDK to create and query a source

namirfa:
Fixes Issue #123 

Created new sets of tests that create a source, and then query it. This is done using the Vantiq.insert() and Vantiq.query() methods, to create and query the source respectively. The Vantiq.delete() method is also used to remove the source once testing is complete.

The tests create a source named "TestObjectRecognitionSource", and there is a note in the README to warn users against creating a source with the same name when building the project.

While writing the tests, I noticed that the NoProcessor does not have any functionality to implement the "savedResolution" configuration options for both source configuration, and query configuration. This would be a simple thing to add, but I am not sure if it is necessary. 

@fhcarter, if you think this is worth adding, I will create a new issue and add it, along with the appropriate tests.

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -298,12 +298,10 @@ it will be set to the default value, "processNextFrame".
             manner: "yyyy-MM-dd--HH-mm-ss".
                 *   To select all images *before* or *after* a certain date, the "-" value can be used as one of the date 
                 strings in the list. For example, the following would save all dates *before* the given date: 
-                    *   \["-", yourEndDate\]. 
-                *   To select *all files* in the output directory, one must use the following value for imageDate: 
-                    *   \["-", "-"\].
-                    *   **NOTE:** This will only select all of the files that are named according to the default naming 
-                    pattern, ("yyyy-MM-dd--HH-mm-ss"). The user is responsible for uploading any files with unique names, and 
-                    can accomplish this using the imageName option.
+                    *   \["-", yourEndDate\]
+                *   To select *all files named with the date-naming convention* in the output directory, one must use the 
+                following value for imageDate: 
+                    *   \["-", "-"\]
         *   "savedResolution": Optional. This value can be set in the same way as it is set in the source configuration. If it 
         is defined here as a query parameter, it will override the value set in the source configuration, otherwise the source 
         configuration value will be used. The setting cannot be larger than that provided in the source configuration (since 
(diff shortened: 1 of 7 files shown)


objectRecognitionSource/README.md
```diff
@@ -322,6 +325,9 @@ it will be set to the default value, "processNextFrame".
                     *   \["-", yourEndDate\]. 
                 *   To select *all files* in the output directory, one must use the following value for imageDate:
```
fhcarter: similar comment

objectRecognitionSource/README.md
```diff
@@ -673,6 +679,9 @@ OpenCV will also display errors that look like `warning: Error opening file
 (/build/opencv/modules/videoio/src/cap_ffmpeg_impl.hpp:857)`. These are expected, as the tests need to ensure correct
 behavior when the file cannot be found.
 
+Some sets of tests will create a Source named "TestObjectRecognitionSource" in the VANTIQ Namespace associated with the
```
fhcarter: Would it be worth it to provide an override -- allow a gradle property to set this -- otherwise, we'll use this name?

Might be good to name it something less likely to exist -- UnlikelyToExistTestObjectRecognizer or something like that, as _TestObjectRecognizer_ might be the first thing anyone tries.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestNoProcessorQueries.java
```diff
@@ -0,0 +1,326 @@
...
+        objRecConfig.put("general", general);
+        objRecConfig.put("neuralNet", neuralNet);
+        
+        // Putting objRecConfig in the source configuration
+        sourceConfig.put("objRecConfig", objRecConfig);
+        
+        // Setting up the source definition
+        sourceDef.put("config", sourceConfig);
+        sourceDef.put("name", SOURCE_NAME);
+        sourceDef.put("type", "ObjectRecognition");
```
fhcarter: Should check that this source implementation exists & complain if it's not there.  And probably put a note in the readme that says that we assume it's existence.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloQueries.java
```diff
@@ -0,0 +1,895 @@
...
+            }
+        }
+    }
+    
+    // ================================================= Helper functions =================================================
+    
+    public static void setupSource(Map<String,Object> sourceDef) {
+        VantiqResponse insertResponse = vantiq.insert("system.sources", sourceDef);
+        if (insertResponse.isSuccess()) {
+            core = new ObjectRecognitionCore(SOURCE_NAME, testAuthToken, testVantiqServer, MODEL_DIRECTORY);;
```
fhcarter: As noted, it may make sense to have this NOT be a constant, but have it read in from gradle props/command line or something.

namirfa:
@fhcarter 
I made the changes to the README that you suggested, and I changed the name of the Source to be "UnlikelyToExistTestObjectRecognitionSource". Hopefully that's unlikely enough of a name!

In terms of making the source name configurable via gradle.properties, I could definitely add this. The only thing I am worried about is that I would have to change all of the constants that are dependent on the source name, (namely all of the IMAGE_1 filenames, etc...).

If you still think it's worth doing, I can get that done now! Let me know.

objectRecognitionSource/README.md
rblumer4: It might enough to add a prefix, e.g. VV__TestObjectRecognitionSource

Issue #139, opened 2019-02-28
JDBC: DateTime elements are not returned correctly


namirfa:
When querying for DateTime elements, the data is returned incorrectly. Seems like the SimpleDateFormat being used to parse the retrieved data is bad.

rblumer4:
email about issue:
-----------------
From: Xiaoguo Liu <xliu@vantiq.com>
Sent: Thursday, February 28, 2019 2:08 AM
To: Technical Team; Namir Fawaz
Subject: Reading DateTime from MySQL databse incorrectly using the JDBC connector 
 
Hi,
 
Today, I just tried to read a date time defined in my MySQL database using our JDBC connector. I initialize it in either of the following way in the MySQL database:
 
INSERT INTO system_plan VALUES
("BJ-C1-201902011145", "Plan001", "基础开工", "2019-5-30 00:00:00", "Plan002", "主体结构施工至正负零", "2019-7-30 00:00:00");
 
INSERT INTO system_plan VALUES
("BJ-C1-201902011145", "Plan001", "基础开工", "2019-5-30", "Plan002", "主体结构施工至正负零", "2019-7-30");
 
I checked it correctly shown in the MySql database:
 
mysql> select * from system_plan
    -> ;
+--------------------+-------------+---------------+---------------------+-------------+----------------------+---------------------+
| ID                 | 管控节点ID1 | 管控节点名称1 | 管控节点计划时间1   | 管控节点ID2 | 管控节点名称2        | 管控节点计划时间2   |
+--------------------+-------------+---------------+---------------------+-------------+----------------------+---------------------+
| BJ-C1-201902011145 | Plan001     | 基础开工      | 2019-05-30 00:00:00 | Plan002     | 主体结构施工至正负零 | 2019-07-30 00:00:00 |
+--------------------+-------------+---------------+---------------------+-------------+----------------------+---------------------+
1 row in set (0.00 sec)
 
However, when I read it inside our Vantiq platform and show it in a topic, it is something like:
 
{
  "管控节点名称1": "基础开工",
   "管控节点ID1": "Plan001",
   "管控节点计划时间1": "2019-30-00T08:00:00.000+0800",
   "管控节点ID2": "Plan002",
   "管控节点计划时间2": "2019-30-00T08:00:00.000+0800",
   "管控节点名称2": "主体结构施工至正负零",
   "ID": "BJ-C1-201902011145"
}
 
[...]

PR #140, opened 2019-02-28
Fixed SimpleDateFormat to properly format the date coming from the DB.

namirfa:
Fixes Issue #139 

There was a simple mistake with the SimpleDateFormat, which caused them to misinterpret datetime elements coming from the database.

@fhcarter 
DateTime elements stored in a DB are returned from the JDBC Extension Source as strings. Is there anyway we could handle incoming dates on the server side, and create VANTIQ Date Types implicitly. If this isn't possible that's fine, I've checked that you can use vail to format the incoming date strings into Vantiq Date Types. This would just remove that extra step.

### DIFF
--- a/jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBC.java
+++ b/jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBC.java
@@ -49,9 +49,6 @@ public class TestJDBC extends TestJDBCBase {
     static final String INSERT_NO_FIELD = "INSERT INTO Test VALUES (1, 25, 'Santa', 'Claus', 'jibberish')";
     static final String INSERT_WRONG_TYPE = "INSERT INTO Test VALUES ('string', 'string', 3, 4)";
     
-    DateFormat dfTimestamp  = new SimpleDateFormat("yyyy-dd-mm'T'HH:mm:ss.SSSZ");
-    DateFormat dfDate       = new SimpleDateFormat("yyyy-dd-mm");
-    DateFormat dfTime       = new SimpleDateFormat("HH:mm:ss.SSSZ");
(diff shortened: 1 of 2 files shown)


jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBC.java
```diff
@@ -49,9 +49,6 @@
     static final String INSERT_NO_FIELD = "INSERT INTO Test VALUES (1, 25, 'Santa', 'Claus', 'jibberish')";
     static final String INSERT_WRONG_TYPE = "INSERT INTO Test VALUES ('string', 'string', 3, 4)";
     
-    DateFormat dfTimestamp  = new SimpleDateFormat("yyyy-dd-mm'T'HH:mm:ss.SSSZ");
-    DateFormat dfDate       = new SimpleDateFormat("yyyy-dd-mm");
-    DateFormat dfTime       = new SimpleDateFormat("HH:mm:ss.SSSZ");
     static final String timestampPattern = "\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d{3}-\\d{4}";
     static final String datePattern = "\\d{4}-\\d{2}-\\d{2}";
     static final String timePattern = "\\d{2}:\\d{2}:\\d{2}.\\d{3}-\\d{4}";
```
fhcarter: I think we need to extend the test here.  This test passed before but wasn't correct.  We should track what date we put in & verify it that we get the same thing (or the expected thing) back.  As the test is, I don't think it will verify that the date is correct does it?  I think we just ask if it _looks like a date_.

rblumer4: Yes, it's best to run your test WITHOUT the fix, to make sure the test fails and it catches the problem. If it doesn't fail, then you need to improve the test.

THEN run the test with the fix to make sure it passes with the fix.

PR #141, opened 2019-03-01
Naf jdbc date tests

namirfa:
Added a test to setup a VANTIQ Source and Publish/Query against it using the VANTIQ SDK. Then set up a Type with a DateTime property, and inserted the DateTime element we queried for into that Type. Finally, checked that the DateTime was converted properly, using the timezone offset that was provided.

(In this case, the DateTime property added 7 hours to the value, since the offset is listed as 0700.)

### DIFF
--- a/jdbcSource/build.gradle
+++ b/jdbcSource/build.gradle
@@ -49,10 +49,10 @@ tasks.withType(Test) {
         systemProperty "EntConJDBCURL", rootProject.findProperty("EntConJDBCURL") ?: "empty"
     }
     if (rootProject.hasProperty("EntConTestSourceName")) {
-        systemProperty "EntConTestSourceName", rootProject.findProperty("EntConTestSourceName") ?: "empty"
+        systemProperty "EntConTestSourceName", rootProject.findProperty("EntConTestSourceName")
     }
     if (rootProject.hasProperty("EntConTestTypeName")) {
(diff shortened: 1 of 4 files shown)


namirfa:
I am editing the README now to make sure that the user adds the TestAuthToken and TestVantiqServer values to their gradle.properties file.

namirfa:
@fhcarter 
I added fixes based on your reviewer comments. Let me know how it looks, and if there is anything else I should change!

jdbcSource/build.gradle
```diff
@@ -48,6 +48,12 @@ tasks.withType(Test) {
     if (rootProject.hasProperty("EntConJDBCURL")) {
         systemProperty "EntConJDBCURL", rootProject.findProperty("EntConJDBCURL") ?: "empty"
     }
+    if (rootProject.hasProperty("EntConTestSourceName")) {
+        systemProperty "EntConTestSourceName", rootProject.findProperty("EntConTestSourceName") ?: "empty"
```
fhcarter: What's the purpose of the "empty"?  Seems like we've already checked for existence.  Moreover, in TestJDBCBase.java, lines 31 & 32 (ish), we only check for null.  So if this sets it to empty, then we'll create sources & types named "empty" which isn't, I don't think, the intent.  Seems like there's no deed for a default value here since we've already verified presence -- it won't change in the meantime in this case.

Issue #143, opened 2019-03-05
JDBC connector crashes when datetime field is NULL


liu-xiao-guo:
I have a database like the following:
![database](https://user-images.githubusercontent.com/8455957/53804457-84ee0300-3f82-11e9-8753-04ba694aa2c0.png)

The datetime field is not a compulsory field for the database. Once I get it running, it crashes the JDBC connector with the following error:

der - Map of the received message: {op=configureExtension, resourceName=sources, isSystemResource=true, resourceId=SapOrder, parameters={}, object={mapId=b7a930aa-c681-478e-8fe7-4dae76b200fd, config={jdbcConfig={general={dbURL=jdbc:mysql://localhost:3306/xugong?useSSL=false&serverTimezone=UTC, password=1234, pollQuery=SELECT * FROM test_table, pollTime=3000, username=root}}, vantiq={packageRows=true}}}, sessionId=bb865875b0dd26a7b25822f63c605790, contentType=application/json, skipMonitoring=false, isExternal=false, address=673d6f32-8b96-4a8b-b6b8-2c038a1cabe4, messageHeaders={}}
20:02:57.902 [OkHttp https://dev.vantiq.cn/...] INFO  io.vantiq.extjsdk.ExtensionWebSocketListener#SapOrder - Successful connection to SapOrder
Exception in thread "executePolling" java.lang.NullPointerException
        at java.util.Calendar.setTime(Calendar.java:1770)
        at java.text.SimpleDateFormat.format(SimpleDateFormat.java:943)
        at java.text.SimpleDateFormat.format(SimpleDateFormat.java:936)
        at java.text.DateFormat.format(DateFormat.java:345)
        at io.vantiq.extsrc.jdbcSource.JDBC.createMapFromResults(JDBC.java:134)
        at io.vantiq.extsrc.jdbcSource.JDBC.processQuery(JDBC.java:71)
        at io.vantiq.extsrc.jdbcSource.JDBCCore.executePolling(JDBCCore.java:249)
        at io.vantiq.extsrc.jdbcSource.JDBCHandleConfiguration$3.run(JDBCHandleConfiguration.java:195)
        at java.util.TimerThread.mainLoop(Timer.java:555)
        at java.util.TimerThread.run(Timer.java:505)

liu-xiao-guo:
@namirfa could you please fix the problem as I am doing a project using JDBC. Also, a developer just spotted this problem when he tried to use SQL server.

Thanks

PR #144, opened 2019-03-05
Fixing JDBC Connector bug

namirfa:
Fixes Issue #143 

Fixing bug related to JDBC Connector crashing when date fields in SQL table are null.  Crash resulted from trying to use a SimpleDateFormat on a null value. Added test to ensure the appropriate behavior occurs.

### DIFF
--- a/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
+++ b/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
@@ -119,35 +119,33 @@ public class JDBC {
                         int columnType = md.getColumnType(i);
                         switch (columnType) {
                             case java.sql.Types.DECIMAL:
-                                row.put(md.getColumnName(i), queryResults.getBigDecimal(i));
+                                if (queryResults.getBigDecimal(i) != null) {
+                                    row.put(md.getColumnName(i), queryResults.getBigDecimal(i));
+                                }
                                 break;
                             case java.sql.Types.DATE:
                                 Date rowDate = queryResults.getDate(i);
                                 if (rowDate != null) {
                                     row.put(md.getColumnName(i), dfDate.format(rowDate));
-                                } else {
-                                    row.put(md.getColumnName(i), null);
                                 }
                                 break;
                             case java.sql.Types.TIME:
                                 Time rowTime = queryResults.getTime(i);
                                 if (rowTime != null) {
                                     row.put(md.getColumnName(i), dfTime.format(rowTime));
-                                } else {
-                                    row.put(md.getColumnName(i), null);
                                 }
                                 break;
                             case java.sql.Types.TIMESTAMP:
                                 Timestamp rowTimestamp = queryResults.getTimestamp(i);
                                 if (rowTimestamp != null) {
                                     row.put(md.getColumnName(i), dfTimestamp.format(rowTimestamp));
-                                } else {
-                                    row.put(md.getColumnName(i), null);
                                 }
                                 break;
                             default:
                                 // If none of the initial cases are met, the data will be converted to a String via getObject()
-                                row.put(md.getColumnName(i), queryResults.getObject(i));
+                                if(queryResults.getObject(i) != null) {
+                                    row.put(md.getColumnName(i), queryResults.getObject(i));
+                                }
                                 break;
                         }
                     }
--- a/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
+++ b/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
@@ -123,15 +123,27 @@ public class JDBC {
                                 break;
                             case java.sql.Types.DATE:
                                 Date rowDate = queryResults.getDate(i);
-                                row.put(md.getColumnName(i), dfDate.format(rowDate));
+                                if (rowDate != null) {
+                                    row.put(md.getColumnName(i), dfDate.format(rowDate));
+                                } else {
+                                    row.put(md.getColumnName(i), null);
+                                }
                                 break;
                             case java.sql.Types.TIME:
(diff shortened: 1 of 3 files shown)


jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -123,15 +123,27 @@ public int processPublish(String sqlQuery) throws VantiqSQLException {
                                 break;
                             case java.sql.Types.DATE:
                                 Date rowDate = queryResults.getDate(i);
-                                row.put(md.getColumnName(i), dfDate.format(rowDate));
+                                if (rowDate != null) {
+                                    row.put(md.getColumnName(i), dfDate.format(rowDate));
+                                } else {
+                                    row.put(md.getColumnName(i), null);
+                                }
```
fhcarter: Is there a reason we need the column there?  Since the row's a map, can't we just omit it?

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -123,15 +123,27 @@ public int processPublish(String sqlQuery) throws VantiqSQLException {
...
                                 break;
                             case java.sql.Types.TIMESTAMP:
                                 Timestamp rowTimestamp = queryResults.getTimestamp(i);
-                                row.put(md.getColumnName(i), dfTimestamp.format(rowTimestamp));
+                                if (rowTimestamp != null) {
+                                    row.put(md.getColumnName(i), dfTimestamp.format(rowTimestamp));
+                                } else {
+                                    row.put(md.getColumnName(i), null);
+                                }
```
fhcarter: So, suppose the date returned is invalid somehow or some other exception is thrown?  How are we preventing the overall crash (assuming that's what occurred)?  

That is, if there was a connector crash (or escaped exception that prevent us from leaving things in a sensible state), we need to ensure that we leave ourselves in a sensible state.  We may not be able to run the query in question, but we should be able to continue operation.

namirfa:
@fhcarter 
I just added a "catch(Exception e)" for each of the three methods that interact with JDBC directly:
* executePublish()
* executeQuery()
* executePolling()

For the executeQuery() method, a queryError is sent back to VANTIQ. The other two methods simply log the error as an "unexpected error".

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -253,6 +262,8 @@ public synchronized void executePolling(String pollQuery) {
             }
         } catch (VantiqSQLException e) {
             log.error("Could not execute polling query.", e);
+        } catch (Exception e) {
+            log.error("An unexpected error occured when executing the polling query.");
```
fhcarter: For all of these, we need to include whatever data we need to debug.  So, we should log the error we got (not just "oops, something happened."  Where we can, the error should include the request that precipitated it and the response we got that we failed to deal with.

namirfa:
@fhcarter 
Added fixes based on your comments. Changed the additional test to check null values for all types, not just date types.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -234,6 +240,9 @@ public void executePublish(ExtensionServiceMessage message) {
         } catch (VantiqSQLException e) {
             log.error("Could not execute requested query.", e);
             log.debug("Request was: {}", request);
+        } catch (Exception e) {
+            log.error("An unexpected error occured when executing the requested query.", e);
```
fhcarter: occured -> occurred (in all the exceptions)

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -119,23 +119,33 @@ public int processPublish(String sqlQuery) throws VantiqSQLException {
...
                                 Timestamp rowTimestamp = queryResults.getTimestamp(i);
-                                row.put(md.getColumnName(i), dfTimestamp.format(rowTimestamp));
+                                if (rowTimestamp != null) {
+                                    row.put(md.getColumnName(i), dfTimestamp.format(rowTimestamp));
+                                }
                                 break;
                             default:
                                 // If none of the initial cases are met, the data will be converted to a String via getObject()
-                                row.put(md.getColumnName(i), queryResults.getObject(i));
+                                if(queryResults.getObject(i) != null) {
```
fhcarter: space after 'if'

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -253,6 +262,9 @@ public synchronized void executePolling(String pollQuery) {
             }
         } catch (VantiqSQLException e) {
             log.error("Could not execute polling query.", e);
```
fhcarter: why do we not log the request here?  We do in the publish case.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -234,6 +240,9 @@ public void executePublish(ExtensionServiceMessage message) {
         } catch (VantiqSQLException e) {
             log.error("Could not execute requested query.", e);
             log.debug("Request was: {}", request);
+        } catch (Exception e) {
+            log.error("An unexpected error occured when executing the requested query.", e);
+            log.debug("Request was: {}", request);
```
fhcarter: Do you not want the request and/or the result when an error occurs? That is, why is this debug vs. error?

PR #148, opened 2019-04-04
Initial JMS Source

namirfa:
This is the initial version of the JMS Source.

**Still needs completing:**
- Handling oddball JMS Message Types
- Adding the README
- Adding an example Source Implementation
- Adding an example VANTIQ Project

I have created this PR so that @fhcarter can begin reviewing the codebase while I finish writing the README, etc.


jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageListener.java
```diff
@@ -0,0 +1,140 @@
...
+                msgMap.put("message", msgText);
+                msgMap.put("JMSFormat", TEXT);
+                msgMap.put("destination", this.destName);
+            } else if (message instanceof ObjectMessage) {
+                Object msgObject = ((ObjectMessage) message).getObject();
+                msgMap.put("message", msgObject);
+                msgMap.put("JMSFormat", OBJECT);
+                msgMap.put("destination", this.destName);
+            } else if (message instanceof MapMessage) {
+
```
fhcarter: Assuming these are things still to be done.

FYI:  Convention is generally to use `// FIXME: some description` or just `// FIXME`.  This indicates to the reviewer that you know stuff's missing, that it will be fixed before merge to master, etc.  As it is, things just fail silently which is the worst kind of failure.

For things that we should deal with but may do after merge/release, use `// TODO`.

IDE's recognize these and call them out (they show up as bright yellow for me, and are marked in the scroll bar).

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageProducer.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param dest                      The name of the destination to connect to (topic or queue)
+     * @param isQueue                   A boolean flag used to create the appropriate type of destination (queue or topic)
+     * @param username                  The username used to create the JMS Connection, (or null if JMS Server does not require auth)
+     * @param password                  The password used to create the JMS Connection, (or null if JMS Server does not require auth)
+     * @throws NamingException
+     * @throws JMSException
+     */
+    public void setupMessageProducer(String connectionFactoryName, String dest, boolean isQueue, String username, String password) throws NamingException, JMSException {
+        this.destName = dest;
+        connectionFactory = (ConnectionFactory) context.lookup(connectionFactoryName);
```
fhcarter: same set of comments about error handling as in `JMSMessageListener`

Thus super/sub-class comment.  Naming this `open()` makes that easier...

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageProducer.java
```diff
@@ -0,0 +1,120 @@
...
+                BytesMessage byteMessage = session.createBytesMessage();
+                break;
+            case TEXT:
+                TextMessage textMessage = session.createTextMessage();
+                textMessage.setText(message);
+                producer.send(textMessage);
+                break;
+            case STREAM:
+                // Handle it
+                StreamMessage streamMessage = session.createStreamMessage();
```
fhcarter: ditto `// FIXME`'s

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageProducer.java
```diff
@@ -0,0 +1,120 @@
...
+                break;
+            case MAP:
+                // Handle it
+                MapMessage mapMessage = session.createMapMessage();
+                break;
+            case OBJECT:
+                // Handle it
+                ObjectMessage objectMessage = session.createObjectMessage();
+                break;
+        }
```
fhcarter: need `default:` case for blown software.  Or JMS may evolve or some provider might implement `fredsWonderFormat` which we don't support and should note...

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageProducer.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    
+    /**
+     * A method used to close the JMS Session and Connection
+     * @throws JMSException
+     */
+    public void close() throws JMSException {
```
fhcarter: synchro issues?

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageListener.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+                
+        return msgMap;
+    }
+    
+    /**
+     * A method used to close the JMS Session and Connection
+     * @throws JMSException
+     */
+    public void close() throws JMSException {
```
fhcarter: synchro

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageProducer.java
```diff
@@ -0,0 +1,120 @@
...
+import javax.jms.Message;
+import javax.jms.MessageProducer;
+import javax.jms.ObjectMessage;
+import javax.jms.Session;
+import javax.jms.StreamMessage;
+import javax.jms.TextMessage;
+import javax.naming.Context;
+import javax.naming.NamingException;
+
+public class JMSMessageProducer {
```
fhcarter: is there enough commonality to build a superclass that handles `close()` and a bunch of the error handling in `setup...()`?

For JMSMessage{Listener,Producer}, should `setupWhatever()` be renamed to `open()` since that's the opposite of `close()` and it actually opens/enables/connects/whatever the `Listener`/`Producer`?

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSQueueMessageConsumer.java
```diff
@@ -0,0 +1,121 @@
...
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.ObjectMessage;
+import javax.jms.Session;
+import javax.jms.StreamMessage;
+import javax.jms.TextMessage;
+import javax.naming.Context;
+import javax.naming.NamingException;
+
+public class JMSQueueMessageConsumer {
```
fhcarter: ditto super/sub

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSQueueMessageConsumer.java
```diff
@@ -0,0 +1,121 @@
...
+        connectionFactory = (ConnectionFactory) context.lookup(connectionFactoryName);
+        connection = connectionFactory.createConnection(username, password);
+        session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+        destination = session.createQueue(queue);
+        consumer = session.createConsumer(destination);
+        connection.start();
+    }
+    
+    /**
+     * Called by the JMS Class, and used to read the most recent JMS Message from the associated queue
```
fhcarter: "recent"...

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSQueueMessageConsumer.java
```diff
@@ -0,0 +1,121 @@
...
+            msgMap.put("message", msgText);
+            msgMap.put("JMSFormat", TEXT);
+            msgMap.put("destination", this.destName);
+        } else if (message instanceof ObjectMessage) {
+            Object msgObject = ((ObjectMessage) message).getObject();
+            msgMap.put("message", msgObject);
+            msgMap.put("JMSFormat", OBJECT);
+            msgMap.put("destination", this.destName);
+        } else if (message instanceof MapMessage) {
+
```
fhcarter: FIXME's

probably good practice here (and other places with similar content) to throw exception or otherwise fail spectacularly when called with things it cannot do.  FIXME says I'll fix it, but don't want things to slip out that fail silently.  

throw SomeNotYetSupported('not done yet') type of thing...

jmsSource/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,15 @@
...
+    <Appenders>
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+        </Console>
+    </Appenders>
+    <Loggers>
+        <Logger name="io.vantiq.extsrc.jmsSource" level="debug"/>
+        <Logger name="io.vantiq.extjsdk" level="debug"/>
+        <Root level="error">
+            <AppenderRef ref="Console"/>
```
fhcarter: To help "some folks", would it make sense to add some appender here that writes to a file that it truncates on each start?  Then, there'd be a record.  

Maybe define it but comment it out, making it easier for anyone to use it.

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/TestJMS.java
```diff
@@ -0,0 +1,270 @@
...
+
+import javax.jms.JMSException;
+import javax.naming.NamingException;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.Test;
+
+import io.vantiq.client.SubscriptionCallback;
```
fhcarter: These seem to be used only in tests, so we should make the changes in the `build.gradle` file, I think.

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/TestJMS.java
```diff
@@ -0,0 +1,270 @@
...
+        // Close JMS if still open
+        if (jms != null) {
+            jms.close();
+            jms = null;
+        }
+    }
+    
+    @Test
+    public void testProduceAndConsumeQueueMessage() {
+        assumeTrue(jmsDriverLoc != null && testJMSURL != null && testJMSConnectionFactory != null && testJMSInitialContext != null
```
fhcarter: since this is used all over, can you move the `assumeTrue` to some method and just call it?  Then easier to change...

jmsSource/build.gradle
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Copies the logger setup into logConfig in the distribution
+applicationDistribution.from("src/main/resources") {
+    include "log4j2.xml"
+    into "logConfig"
+}
+
+tasks.withType(Test) {
+    if (System.env.JMS_DRIVER_LOC == null) {
```
fhcarter: do we have a mechanism to run tests against a variety of these?  Some über script that rotates through our test subjects?  May require some machine with them all set up?

jmsSource/build.gradle
```diff
@@ -0,0 +1,96 @@
+group 'io.vantiq'
```
fhcarter: Could we make this build log a message but not do anything if no JMS_DRIVER_LOC defined?

namirfa:
@fhcarter 
Added fixes for all of the reviewer comments, and handled the oddball JMS Types. I am working now on getting the README finished, and once this is complete I will and an example VANTIQ Project.

One thing to note, I am aware that there is repeated code when checking the MessageHandlerInterface implementation, (for all the message consumers/listeners/producers). I thought it might be a bit harder to follow the code if I pulled this out into a separate utility class/function, but if you disagree then feel free to let me know and I can easily change this!

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/TestJMS.java
fhcarter: `checkAllJMSProperties()` or whatever it's called?

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMS.java
```diff
@@ -0,0 +1,278 @@
...
+            }
+            
+            msgProducer.produceMessage(message, messageFormat);
+        }
+    }
+    
+    /**
+     * A method used to close all of the resources being used by the message producers/consumers/listeners
+     */
+    public void close() {
```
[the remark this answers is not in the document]
namirfa: We mark JMS as closed by setting it to null (spoke about this in person)

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSHandleConfiguration.java
```diff
@@ -0,0 +1,286 @@
...
+
+    Handler<ExtensionServiceMessage> queryHandler;
+    Handler<ExtensionServiceMessage> publishHandler;
+
+    public JMSHandleConfiguration(JMSCore source) {
+        this.source = source;
+        this.sourceName = source.getSourceName();
+        log = LoggerFactory.getLogger(this.getClass().getCanonicalName() + "#" + sourceName);
+        queryHandler = new Handler<ExtensionServiceMessage>() {
+            ExtensionWebSocketClient client = source.client;
```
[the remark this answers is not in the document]
namirfa: Spoke about this in person

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMS.java
```diff
@@ -0,0 +1,278 @@
...
+ * JMS Message Producers/Consumers/Listeners. Initialized and setup according to the source configuration, which specifies
+ * the queues and/or topics to connect to, as well as the connection factory to use when connecting.
+ */
+public class JMS {
+    
+    private static final String MESSAGE_HANDLER_FQCN = "io.vantiq.extsrc.jmsSource.communication.messageHandler.DefaultMessageHandler";
+    private static final String SYNCH_KEY = "synchKey";
+    
+    Logger log  = LoggerFactory.getLogger(this.getClass().getCanonicalName());
+    ExtensionWebSocketClient client;
```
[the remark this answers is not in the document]
namirfa: Also spoke about this one.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSHandleConfiguration.java
```diff
@@ -0,0 +1,286 @@
...
+
+            @Override
+            public void handleMessage(ExtensionServiceMessage message) {
+                String replyAddress = ExtensionServiceMessage.extractReplyAddress(message);
+                
+                // Should never happen, but just in case something changes in the backend
+                if ( !(message.getObject() instanceof Map) ) {
+                    client.sendQueryError(replyAddress, "io.vantiq.extsrc.JMSHandleConfiguration.invalidQueryRequest",
```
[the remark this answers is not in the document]
namirfa: Also spoke about this one.

namirfa:
@fhcarter 

I added fixes to your reviewer comments. I also left comments on the ones we spoke about in person. Assuming this all looks good, I'll work on finishing up the README and adding a sample project.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSCore.java
```diff
@@ -0,0 +1,344 @@
...
+                      "JMS connection closed before operation could complete.", null);
+          } else {
+              log.error("JMS connection closed before operation could complete.");
+          }
+      } else {
+          // Retrieve most recent message from specified queue. If no queue name is specified, or if exception is thrown, return query error.
+          if (request.get("queue") instanceof String) {
+              String queue = (String) request.get("queue");
+              try {
+                  Map<String, Object> messageMap = localJMS.consumeMessage(queue);
```
rblumer4: Fred, Namir,
localJMS could be closed here, since close() could have been called after the SYNCH block above.
I guess one of the catch() clauses below will handle that case?

namirfa: This is indeed possible. We catch all of the exceptions that we expect to encounter, but also have a ```catch (Exception e)``` block at the end in case something unexpected happens. 

I think this could be considered an "unexpected" error, since it is pretty unlikely that the JMS instance would be closed between checking that it is non-null and actually using it to consume a queue message. Let me know if this is an inaccurate assumption.

fhcarter: @namirfa I think it's expected in the sense that it's a documented exception that could come back.  Someone well could close it.  However, I think you're correct that the, to answer @rblumer4's question, we understand that that's quite possible, and the expectation is that we'll catch the exception.

The alternative, as you know, was to block access during the entirety of any JMS operation.  And since those may involve network or other/separate server access, that seems like a bad idea.

rblumer4: that sound ok to me, too.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSHandleConfiguration.java
```diff
@@ -0,0 +1,297 @@
...
+            return false;
+        }
+
+        if (generalConfig.get("initialContext") instanceof String) {
+            initialContext = (String) generalConfig.get("initialContext");
+            
+            // Check to make sure we only have one InitialContextFactory
+            String initialContextCheck = source.initialContextMap.get(initialContext);
+            if (initialContextCheck == null) {
+                if (source.initialContextMap.size() > 1) {
```
fhcarter: I think the test here should be
```
if (source.initialContextMap.size() > 0) {
```

( or `!= 0`). That is, we don't have one with this name already, so we're going to need to add it.  Once we add it, will we be over the limit of "1" -- meaning we were at zero before that.

Even easier might be change the map to a string, and if it's null or matches the requested one, then things are OK. Otherwise, not.

namirfa: I am going to change it to a string since that definitely seems like the simplest solution here, sorry for the silly mistake.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSCore.java
```diff
@@ -0,0 +1,344 @@
...
+            } catch (JMSException e) {
+                log.error("An error occured when attempting to send the given message.", e);
+            } catch (DestinationNotConfiguredException e) {
+                log.error("An error occured when attempting to send the given message. The source was not configured "
+                        + "to send messages to the following destination: " + dest + ".", e);
+            } catch(UnsupportedJMSMessageTypeException e) {
+                log.error("An error occured when attempting to send the given message. The provided JMS Message Type: "
+                        + e.getMessage() + " is either invalid or not currently supported.");
+            } catch (Exception e) {
+                log.error("An unexpected error occured when attempting to send the given message.", e);
```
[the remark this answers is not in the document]
namirfa: We throw a DestinationNotConfiguredException if we could not find a message producer/consumer for a given queue or topic. In the case that Roger was referring to earlier, I believe this would throw a null pointer exception which would be caught in the last catch block. 
What we could do there is check that ```localJMS == null```, and log/send a more specific error message in that case. Would that work?

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+format:
+```
+{
+    JMSMessage:<theReceivedMessage>, 
+    JMSFormat:<theJMSMessageFormat>, 
+    JMSDestination:<queueName/topicName>
+}
+```
+
+The following example shows a rule that could be used to process incoming notifications to a JMS Source:
```
fhcarter: identify that the source's name is `JMS1`

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+var myObj = {}
+myObj.JMSMessage = msg.message
+myObj.JMSFormat = msg.JMSFormat
+myObj.JMSDestination = msg.destination
+INSERT JMSMessageType(myObj)
+```
+
+## Select Statements
+
+In order to read messages from a queue, (**NOT** a queueListener), a VAIL Select statement must be used. The Select statement
```
fhcarter: Select --> SELECT

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+      myObj.JMSDestination = msg.destination
+      INSERT JMSMessageType(myObj)
+  }
+```
+
+*   **NOTE**: Messages returned by a Select statement are in the exact same format as those sent to the source as a Notification.
+
+## Publish Statements
+
+In order to send messages to either a queue or a topic, a VAIL Publish statement must be used. The Publish statement will have
```
fhcarter: Publish --> PUBLISH

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+three parameters: *message*, *queue/topic*, *JMSFormat*. The *message* parameter is simply the message to be sent. The next
+parameter is either *queue* or *topic*, depending on which type of destination you are sending the message to. The value of this
+parameter is the name of the given queue/topic. Finally, the last parameter is the *JMSFormat*, which specifies the JMS Message
+Type that will be used to send the message. If no custom Message Handler was implemented, the three values that we currently
+support are "TextMessage", "MapMessage", and "Message". The "Message" type does not have any message body, so no message parameter
```
fhcarter: ~currently~

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+is needed. The following two examples show how to send all three supported message types.
```
fhcarter: two --> three

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+### Sending Message ###
+```
+PROCEDURE sendMapMessage()
+
+var dest = "NamirJMSServer-0/NamirSystemModule-0!NamirJMSServer-0@/com/namir/weblogic/base/dq"
+var msgFormat = "Message"
+
+PUBLISH {queue: dest, JMSFormat: msgFormat} to SOURCE JMS1
```
fhcarter: Is this, effectively, just a PING-style message?  What are these used for.

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+PROCEDURE sendMapMessage()
+
+var dest = "NamirJMSServer-0/NamirSystemModule-0!NamirJMSServer-0@/com/namir/weblogic/base/dq"
+var msgFormat = "Message"
+
+PUBLISH {queue: dest, JMSFormat: msgFormat} to SOURCE JMS1 
+```
+
+## Custom Message Handlers ##
+
```
fhcarter: Reiterate purpose of message handlers

We may also want to note that this section is for those for developers of message handlers, not configurers of sources.  We might want to have some overview section at the top to identify these roles?

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+To create custom Message Handler, you must implement the 
+```io.vantiq.extsrc.jmsSource.communication.messageHandler.MessageHandlerInterface.java``` interface. We recommend creating a
```
fhcarter: 3 ticks to 1

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+subclass of the BaseMessageHandler class, and adding extra message formatting for the JMS Message Types that we do not 
+currently support. Once you have implemented the MessageHandlerInterface, you must add the class to the java classpath. If we 
+cannot find the class in the java classpath, or if it is not a correct implementation of the interface, then we will not be
+able to create a message producer/consumer/listener for the given queue or topic.
```
fhcarter: will we revert to using our generic handler? or will the source fail to start?

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+    initializes the queryHandler and publishHandler.
+*   **JMS** -- The class that directly interacts with the JMS Server, executing the query and publish requests as sent
+    by the VANTIQ Modelo IDE. Uses classes in the ```io.vantiq.extsrc.jmsSource.communication``` package to produce and 
+    consume messages to or from the JMS Queues and Topics.
+
+## How to Run the Program
+
+1.  In order to effectively use the JMS Source, you will need to download the appropriate JMS Driver with respect to the
+    JMS Server that you are using. Once you have installed this, you will need to create an environment variable named
+    `JMS_DRIVER_LOC` that contains the location of the jar file (*i.e.* `/Users/yourName/somePath/wlthint3client.jar`)
```
fhcarter: Once this is in (or nearly so), I'll add something in the overview readme about the new way that the settings.gradle file works.

jmsSource/REAMDE.md
```diff
@@ -0,0 +1,321 @@
...
+
+PUBLISH {queue: dest, JMSFormat: msgFormat} to SOURCE JMS1 
+```
+
+## Custom Message Handlers ##
+
+To create custom Message Handler, you must implement the 
+```io.vantiq.extsrc.jmsSource.communication.messageHandler.MessageHandlerInterface.java``` interface. We recommend creating a
+subclass of the BaseMessageHandler class, and adding extra message formatting for the JMS Message Types that we do not 
+currently support. Once you have implemented the MessageHandlerInterface, you must add the class to the java classpath. If we
```
fhcarter: You will also need to ensure that any other classes used by your MessageHandler are available on the class path.

jmsSource/README.md
```diff
@@ -0,0 +1,343 @@
...
+## Messages from the Source
+
+Messages that are sent from the JMS Server to a source as VANTIQ Notifications, (from either a queueListener or a topic) are 
+JSON objects in the following format:
+```
+{
+    JMSMessage:<theReceivedMessage>, 
+    JMSFormat:<theJMSMessageFormat>, 
+    JMSDestination:<queueName/topicName>
+}
```
fhcarter: From your example this does not appear to be true.  Your example (below) says that messages arrive as JSON like

```
{
    message:<theReceivedMessage>,
    JMSFormat:<theJMSMessageFormat>,
    destination:<queueName/topicName>
}
```
And that seems to match what's in the code as well.  Seems like we should either use JMS* or not, but the `JMSFormat` sort of sticks out.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSMessageListener.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+    
+    /**
+     * The method used to handle incoming messages. Sends the formatted message to VANTIQ as a Source Notification.
+     */
+    @Override
+    public void onMessage(Message msg) {
+        try {
+            Map<String, Object> msgMap = messageHandler.parseIncomingMessage(msg, destName);
+            client.sendNotification(msgMap);
```
fhcarter: Is there anything here in terms of required properties necessary in the returned `msgMap`?  the format or anything?  If so, we should check them here & document what's required.

From the code, it looks like we expect `message` (possibly optional), `JMSFormat`, and `destination`, and this matches the example code (not the text) in the readme.

Before, returning, we should verify that we go what we expected -- otherwise, complain.

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/TestJMS.java
```diff
@@ -0,0 +1,508 @@
...
+        jms = new JMS(null, testJMSConnectionFactory);
+        try {
+            jms.setupInitialContext(testJMSInitialContext, testJMSURL);
+        } catch (NamingException e) {
+            fail("Should not throw a NamingException when setting up JMS Context. " + e.getMessage());
+        }
+        
+        try {
+            jms.createProducersAndConsumers(sender, receiver, null, null);
+        } catch (Exception e) {
```
fhcarter: Seems like we should verify here that something happened?  Did we log an error or something?  What do we expect?

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/NullMessageHandler.java
```diff
@@ -0,0 +1,22 @@
+package io.vantiq.extsrc.jmsSource;
+
+import java.util.Map;
+
+import javax.jms.Message;
+import javax.jms.Session;
+
+import io.vantiq.extsrc.jmsSource.communication.messageHandler.MessageHandlerInterface;
+
+public class NullMessageHandler implements MessageHandlerInterface {
```
fhcarter: Seems like we will also need an example of a message handler that doesn't return the required properties and/or doesn't create a valid JMS message

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/messageHandler/MessageHandlerInterface.java
```diff
@@ -0,0 +1,42 @@
...
+     * @throws JMSException
+     * @throws UnsupportedJMSMessageTypeException
+     */
+    Message formatOutgoingMessage(Object message, String messageFormat, Session session) throws Exception;
+    
+    /**
+     * A helper function used to parse the incoming message and format it according to its message type
+     * @param message           The retrieved message from the queue
+     * @param destName          The destiation, (queue or topic), from which the message arrived
+     * @return                  A map containing the message, as well as the queue name and the JMS Message Type
```
fhcarter: Either here, in documentation, or both, need to describe what the map keys required are.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/messageHandler/MessageHandlerInterface.java
```diff
@@ -0,0 +1,42 @@
...
+import io.vantiq.extsrc.jmsSource.exceptions.UnsupportedJMSMessageTypeException;
+
+/**
+ * An interface used to handle incoming and outgoing JMS Messages for all Message Producers/Consumers/Listeners
+ */
+public interface MessageHandlerInterface {
+    
+    /**
+     * A helper function used to format the provided message according to whatever JMS Message Type was specified
+     * @param message           The message to be sent, either a string or a map
```
fhcarter: Would it be easier for the case of String, to have a Map with a key of message in it with the string?  That seems more consistent and allows us to tighten up the interface.

jmsSource/README.md
```diff
@@ -0,0 +1,343 @@
...
+The following example shows a rule that could be used to process incoming notifications to a JMS Source named JMS1:
+
+```
+RULE jmsMessageListener
+
+WHEN MESSAGE ARRIVES FROM JMS1 as msg
+var myObj = {}
+myObj.JMSMessage = msg.message
+myObj.JMSFormat = msg.JMSFormat
+myObj.JMSDestination = msg.destination
```
fhcarter: What about the other infrastructure JMS props?  Things like JMSMessageId, correlationId, replyTo, deliveryMode, expiration, timestamp, priority, redelivered?  Seems like we should at least return these in the map.

Seems like we should probably allow them to be set as well, where appropriate.

fhcarter:
@namirfa Why is replyTo unpredictable?

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/communication/JMSQueueMessageConsumer.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Called by the JMS Class, and used to read the next available JMS Message from the associated queue
+     * @return A map containing the message, as well as the queue name and the JMS Message Type
+     * @throws JMSException
+     * @throws UnsupportedJMSMessageTypeException
+     */
+    public Map<String, Object> consumeMessage() throws Exception {
+        try {
+            Message message = consumer.receive(CONSUME_TIMEOUT);
+            Map<String, Object> msgMap = messageHandler.parseIncomingMessage(message, destName, true);
```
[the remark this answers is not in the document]
namirfa: If we timeout, then the `consumer.receive()` method will return `null`. So I don't think there is any special handling we need, but once this pull request is closed I will add an enhancement issue.

jmsSource/src/test/java/io/vantiq/extsrc/jmsSource/TestJMS.java
```diff
@@ -0,0 +1,640 @@
...
+        VantiqResponse queryResponse = vantiq.query(testSourceName, queryParams);
+        
+        // Should query successfully, and message/properties/headers/queue should all be present
+        assert !queryResponse.hasErrors();
+        JsonObject responseBody = (JsonObject) queryResponse.getBody();
+        String responseMessage = (String) responseBody.get("message").getAsString();
+        assert responseMessage.equals(message);
+        assert responseBody.get("headers") instanceof JsonObject;
+        JsonObject headersObject = responseBody.get("headers").getAsJsonObject();
+        assert testJMSQueue.contains(headersObject.get("JMSReplyTo").getAsString());
```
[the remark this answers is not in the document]
namirfa: The reason I wrote the test like this is because the value returned from `JMSReplyTo` is actually a truncated version of the `testJMSQueue` value, which is the full name used to create the queue and send messages.

So I am checking that the value we get back is contained inside the full name we use to actually create JMS Destinations (queues in this case).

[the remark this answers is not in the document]
fhcarter: Ah, OK.  My misunderstanding

PR #150, opened 2019-04-16
Naf obj polling fix

namirfa:
Added ability to decouple capturing and processing images. We now process images in a separate thread, speeding up the frame rate at which we can capture images.

@fhcarter 
I used a ThreadPoolExecutor as the thread pool, and initialized it with a LinkedBlockingQueue with a max capacity of 10 threads. The max queue-size was chosen completely arbitrarily, and can be changed if it is too small. We can also make this a configurable item. As it stands, the ThreadPoolExecutor uses 5 threads, and allows for up to 10 threads to be queued at any moment. If a thread is idle, it will immediately be killed (this is the 0l parameter). 

All tests pass as expected.


objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -323,11 +326,18 @@ boolean createImageRetriever(Map<String, ?> dataSourceConfig) {
     private boolean prepareCommunication(Map<String, ?> general) {
         int polling = -1; // initializing to an invalid input
         boolean queryable = false;
+        source.pool = new ThreadPoolExecutor(5, 5, 0l, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(10));
```
fhcarter: I think it would be good, before we push to master, to have the a configurable `concurrentAnalysisCount` and `queuedAnalysisCount` (to be the 5 & 10, respectively).  That said, assuming other things are OK, we can publish the branch to Brett & company and let them play with it as is.

That shouldn't be very hard to do.  I would suggest that we support `concurrentAnalysisCount (defaults to 5), and `queuedAnalysisCount` (defaults to 2 x `concurrentAnalysisCount`). We will note (in the README) that each of these increases memory pressure (the total of the two images in memory at a time) and CPU/gCPU cost).

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -54,34 +56,22 @@
...
-    
-    // vars for internal use
     ExtensionWebSocketClient    client      = null;
     NeuralNetInterface          neuralNet   = null;
     SimpleDateFormat            format      = new SimpleDateFormat("yyyy-MM-dd--HH-mm-ss");
     
     public String outputDir;
     public String lastQueryFilename;
     
-    // final vars
```
fhcarter: :-)

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -355,28 +356,40 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
+       NeuralNetInterface localNeuralNet = null;
        try {
-           synchronized (this) {
+           synchronized (SYNCH_KEY) {
                if (neuralNet == null) { // Should only happen when close() runs just before sendDataFromImage()
                    if (client != null) {
                        client.sendQueryError(replyAddress, this.getClass().getPackage().getName() + ".closed",
                                "The source closed mid message", null);
                    }
                    return;
```
fhcarter: would be better to move the client.sendQuery() outside the sync block.  No need to block there.  At some level, it doesn't matter because the source already closed, but better code hygiene to have that outside the synchronized block.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -355,28 +356,40 @@ public void sendDataFromImage(ImageRetrieverResults imageResults, ExtensionServi
...
-               } else {
-                   client.sendQueryResponse(200, replyAddress, results.getResults().toArray(new Map[0]));
-               }
+               localNeuralNet = neuralNet;
+           }
+           
+           NeuralNetResults results;
+           if (localNeuralNet instanceof NeuralNetInterface2) {
+               NeuralNetInterface2 localNeuralNet2 = (NeuralNetInterface2) localNeuralNet;
+               results = localNeuralNet2.processImage(processingParams, image, request);
```
fhcarter: here & above, since the local variable is used only on this line, casting is probably a better choice.  The compiler probably figured that out and did it anyway, but clearer what's going on.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
fhcarter: Not for this merge, but may make sense (_i.e._ enter an issue...) to allow a choice about what to happen when the queue gets full.  Should we just drop things (which is what we do now), or should we bias for more recent executions (make it a stack kinda thing, but not really).  This may not be that important because if you're running into queuing issues, you've underallocated/oversubscribed your connector.  But worth considering (we can always decide, "no.")

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
```diff
@@ -230,17 +234,41 @@ private void setup(Map<String, ?> neuralNet, String sourceName, String modelDire
...
+    }
+    
+    /**
+     * Run the image through a YOLO net. May save the resulting image depending on the settings.
+     */
+    @Override
+    public NeuralNetResults processImage(Map<String, ?> processingParams, byte[] image) throws ImageProcessingException {
+        Date timestamp;
+        if (processingParams != null) {
+            timestamp = (Date) processingParams.get("timestamp");
```
fhcarter: processingParams.timestamp --> processingParams.imageTimestamp

Who knows what other timestamps we'll have in the future.

fhcarter: Use constant alluded to above.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/NeuralNetInterface2.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+import io.vantiq.extsrc.objectRecognition.exception.FatalImageException;
+import io.vantiq.extsrc.objectRecognition.exception.ImageProcessingException;
+
+/**
+ * An interface for the neural net that will process the image and return a List of data representing the objects found.
+ */
+public interface NeuralNetInterface2 extends NeuralNetInterface {
+    /**
```
fhcarter: Define constant for IMAGE_TIMESTAMP `public static final String IMAGE_TIMESTAMP = imageTimestamp;`.

Avoid "magic strings" in code.  Also tells implementors of the interface what they're expected to deal with/do.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
fhcarter: This doesn't verify that `IMAGE_TIMESTAMP` exists -- only that some `Map` was passed in.  Either change the test to include that test, or handle its being null.  I don't think you want/expect `timestamp` to be null.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/YoloProcessor.java
```diff
@@ -264,14 +292,22 @@ public NeuralNetResults processImage(byte[] image) throws ImageProcessingExcepti
...
         Vantiq vantiq = null;
         
+        Date timestamp;
+        if (processingParams != null) {
+            timestamp = (Date) processingParams.get("timestamp");
+        } else {
+            timestamp = new Date();
+        }
+        fileName = format.format(timestamp);
+
```
fhcarter: same comments about timestamp handling

namirfa:
@fhcarter 
I added fixes to your comments, except for the first ones regarding the configurable thread count etc.

If you think this looks good enough for Brett and Pat to test it out, I will send them an email. Once we get their feedback, I can make the max thread count configurable, as per your remarks.

Issue #152, opened 2019-04-25
JMS Source: Make Timeout for Queue Message Consumption a Configurable Option


namirfa:
The `JMSQueueMessageReceiver` class uses a constant value to specify the timeout when reading a queue message, (which is currently fixed at one second).

This enhancement will enable a user to optionally configure this timeout value from the Source Configuration. Otherwise, it will be set to the default value of one second.

PR #153, opened 2019-04-25
Making Timeout for Queue Message Consumption an Optional Query Parameter

namirfa:
Closes Issue #152 

Changed the default behavior of SELECTING from the VANTIQ Source to use the `consumer.receiveNoWait()` method. If no `timeout` parameter is specified in the SELECT statement, or if the value is invalid, (not a non-negative integer), then the new default behavior will be used. However, if a valid `timeout` is specified, then it will be used.

### DIFF
--- a/jmsSource/README.md
+++ b/jmsSource/README.md
@@ -241,9 +241,13 @@ INSERT JMSMessageType(myObj)
 ### Select Statements <a name="select" id="select"></a>
 
 In order to read messages from a queue, (**NOT** a queueListener), a VAIL SELECT statement must be used. The SELECT statement 
-must have two query parameters: `operation` and `queue`. Currently there is only one `operation` that is supported, which is 
-the "read" `operation`. In the future, there may be different SELECT Operations. The `queue` parameter is the name of the 
-queue from which to read. The following is an example of a Procedure created in VANTIQ Modelo querying against a JMS Source:
+must have two query parameters: `operation` and `queue`. A third optional parameter, `timeout`, can be specified as well. 
+Currently there is only one `operation` that is supported, which is the "read" `operation`. In the future, there may be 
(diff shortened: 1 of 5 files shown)


jmsSource/README.md
```diff
@@ -241,9 +241,13 @@ INSERT JMSMessageType(myObj)
...
 
 In order to read messages from a queue, (**NOT** a queueListener), a VAIL SELECT statement must be used. The SELECT statement 
-must have two query parameters: `operation` and `queue`. Currently there is only one `operation` that is supported, which is 
-the "read" `operation`. In the future, there may be different SELECT Operations. The `queue` parameter is the name of the 
-queue from which to read. The following is an example of a Procedure created in VANTIQ Modelo querying against a JMS Source:
+must have two query parameters: `operation` and `queue`. A third optional parameter, `timeout`, can be specified as well. 
+Currently there is only one `operation` that is supported, which is the "read" `operation`. In the future, there may be 
+different SELECT Operations. The `queue` parameter is the name of the queue from which to read. The `timeout` parameter can be 
+used to force the queue message consumer to wait up to `timeout` milliseconds for a queue message to arrive. The `timeout` 
+parameter must be a **non-negative integer**, otherwise it will be ignored. If the `timeout` value is set to 0, this will make
```
fhcarter: We might want to consider treating `0` the same as missing.  That is, require a positive integer and don't allow indefinite wait.  Really indefinite won't work anyway -- the user can get around it by waiting for an hour (which will fail when the procedure dies).

namirfa: Sounds good, I'll change this to not allow 0.

namirfa:
@fhcarter 

Changed the code to treat 0 as an invalid `timeout` value.

Issue #158, opened 2019-05-02
JDBC extension source doesn't attempt to reconnect to DB after connection lost


msilva123:
Not sure how to tell the version of the JDBC source, it is not in the Readme, though I downloaded yesterday and built it.

Running on Windows.

1. Built the JDBC Source and configured namespace to work with a local MySQL database.
2. While running, performed a script to query the database - works fine.
3. Restarted the MySQL database - JDBC Source process continues to run
4. Re-ran the same script to query the database - got the following error:
![image](https://user-images.githubusercontent.com/3606537/57051180-1c936900-6c35-11e9-9a3c-fae10f66c2f2.png)

Here is the log from the windows shell:
C:\Users\msilva\Documents\Vantiq\Modeling\MySQL>jdbcSource\bin\jdbcSource
16:29:53.914 [OkHttp https://dev.vantiq.com/...] INFO  io.vantiq.extjsdk.ExtensionWebSocketListener#JDBC1 - WebSocket op
en
16:29:54.218 [OkHttp https://dev.vantiq.com/...] DEBUG io.vantiq.extjsdk.ExtensionWebSocketListener#JDBC1 - Map of the r
eceived message: {status=200, contentType=application/json, body={anonymous=false, userInfo={accessToken=zWp1qIozWR6gqIn
qHWHmfAHVO6iXeewoSo3C5cK0pcI=, idToken=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJ3YXRlcmJpdF9fd2F0ZXJiaXQiLCJwcmVm
ZXJyZWRVc2VybmFtZSI6IndhdGVyYml0X193YXRlcmJpdCIsImlzcyI6Imh0dHBzOi8vdmFudGlxLmNvbS9hdXRoZW50aWNhdGUiLCJwcm9maWxlcyI6WyJX
YXRlckJpdC5hZG1pbiIsIldhdGVyQml0LmFkbWluX19zeXN0ZW0iXSwicHJlZmVycmVkX3VzZXJuYW1lIjoid2F0ZXJiaXRfX3dhdGVyYml0IiwiaG9tZU5h
bWVzcGFjZSI6IldhdGVyQml0IiwiaWRlbnRpdHlQcm92aWRlciI6IlZhbnRpcSIsImF1ZCI6InZhbnRpcS1jbGllbnQiLCJjcmVhdGVkQnkiOiIwNzRhYTNj
[...]

PR #159, opened 2019-05-02
Naf jdbc #158

namirfa:
Closes Issue #158 

Added a `diagnoseConnection` method that checks if the connection is still valid, and tries to reconnect if it is not.

I did not add any tests to check this behavior, since I am not sure how to stop/start the SQL Server from a test. I manually tested this extensively and it works as I expect.

Also removed the global `rs` and `stmt` variables and deleted their close functions. This is because we always create them locally using a "try with resources", which will close the variables once execution is complete (in the same way that a `finally` block would).


namirfa:
@fhcarter 
Adding test we spoke about now, I'll let you know when it's complete.

namirfa:
@fhcarter 
A test has been added to check the behavior.

Issue #163, opened 2019-05-13
Increase message size for the JDBC Connector


msilva123:
Can we increase the message size for the JDBC Connector.

I get the following error stating max message size is 262144 (assuming bytes):
 
17:08:30.598 [OkHttp https://dev.vantiq.com/...] ERROR io.vantiq.extjsdk.ExtensionWebSocketListener#JDBC1 - Map of the r
eceived message: {status=500, contentType=application/json, body=[{code=io.vantiq.server.error, message=Cannot process b
inary frame of size 369122, it would cause message buffer (size 0) to overflow max message size of 262144, params=[], de
tails=java.lang.IllegalStateException: Cannot process binary frame of size 369122, it would cause message buffer (size 0
) to overflow max message size of 262144
        at io.vertx.core.http.impl.WebSocketImplBase$FrameAggregator.handleBinaryFrame(WebSocketImplBase.java:371)
        at io.vertx.core.http.impl.WebSocketImplBase$FrameAggregator.handle(WebSocketImplBase.java:323)
        at io.vertx.core.http.impl.WebSocketImplBase$FrameAggregator.handle(WebSocketImplBase.java:309)
        at io.vertx.core.http.impl.WebSocketImplBase.handleFrame(WebSocketImplBase.java:302)
        at io.vertx.core.http.impl.Http1xServerConnection.handleOther(Http1xServerConnection.java:208)
        at io.vertx.core.http.impl.Http1xServerConnection.handleMessage(Http1xServerConnection.java:141)
        at io.vertx.core.net.impl.ConnectionBase.handleRead(ConnectionBase.java:397)
        at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:320)
        at io.vertx.core.impl.EventLoopContext.execute(EventLoopContext.java:43)
        at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:188)
        at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:174)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
[...]

PR #164, opened 2019-05-13
Naf jdbc message size

namirfa:
Fixes Issue #163 

Changed the JDBC Connector to send messages row by row, so we should never run into the issue of exceeding the maximum message size. Added a test to ensure that this works as expected.

### DIFF
--- a/jdbcSource/README.md
+++ b/jdbcSource/README.md
@@ -149,14 +149,19 @@ will be sent as a unique Notification.
 
 In order to interact with the JDBC Source, one option is to use VAIL to select from the source. To do this, you will need 
 to specify the SQL Query you wish to execute against your database as part of the WITH clause. The SQL Queries used here must 
-only be **SELECT STATEMENTS**. The data will be returned as an array of the resulting rows. A query parameter named 
-`bundleFactor` can be used to specify how many rows of data a given message will contain. This parameter is useful when 
-querying a very large number of rows, where the data is likely to exceed the maximum message size allowed by the VANTIQ 
-System. Specifying this parameter as part of the WITH clause allows the connector to stream the data using multiple messages, 
-each containing `bundleFactor` number of rows. If no `bundleFactor` is specified, then the default value of 1000 will be used. 
-If a `bundleFactor` of 0 is specified, then the connector will attempt to send all of the queried rows in a single message, 
-(**NOTE**: this can likely cause errors if the combined data exceeds the maximum message size). The following is an example of 
-a Procedure created in VANTIQ Modelo querying against a JDBC Source.
+only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message contains some 
+number of rows.
+
+A query parameter named `bundleFactor` determines how many rows are bundled into each message. Generally, the default value of 
+500 will be fine. However, in cases where rows may be very large, a smaller value may be required.
+
+The `bundleFactor` parameter is specified in the `WITH` clause, allowing each query to adjust as appropriate. If the parameter 
+is not specified, the default value (500) is used -- that will be fine for most cases. If a value of `0` is provided, then all 
+rows will be placed into a single message (the use of this is discouraged as it may cause problems). If the value is positive 
+(and non-zero), then that many rows will be sent at a time.
+
+From the perspective of consuming the rows, there is no visible difference here. The `bundleFactor` parameter is present to 
+allow control when returning very large rows.
 
 ```
 PROCEDURE queryJDBC()
--- a/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
+++ b/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
@@ -278,7 +278,7 @@ public class JDBCCore {
     */
    public void sendDataFromQuery(HashMap[] queryArray, ExtensionServiceMessage message) {
        String replyAddress = ExtensionServiceMessage.extractReplyAddress(message);
-              
+       
        // Send the results of the query
        if (queryArray.length == 0) {
            // If data is empty send empty map with 204 code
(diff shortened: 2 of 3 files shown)


jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -33,21 +34,22 @@
...
-    // vars for internal use
     Timer                       pollTimer = null;
     ExtensionWebSocketClient    client  = null;
     JDBC                        jdbc    = null;
     
-    // final vars
     final Logger log;
-    final static int    RECONNECT_INTERVAL = 5000;
+    final static int RECONNECT_INTERVAL = 5000;
+    final static int DEFAULT_BUNDLE_SIZE = 500;
```
fhcarter: query limit (albeit not quite the same thing) is 1000 for VANTIQ queries when returning to an array.  That value got Mark into trouble, so might want to leave it as is -- just noting a consistency issue, sort of.

namirfa:
@fhcarter 

Just added the revisions we spoke about.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -277,14 +279,45 @@ public synchronized void executePolling(String pollQuery) {
...
+       } else if (bundleFactor == 0) {
+           // If the bundleFactor was specified to be 0, then we sent the entire array
            client.sendQueryResponse(200, replyAddress, queryArray);
+           lastRowBundle = queryArray;
+       } else {
+           // Otherwise, send messages containing 'bundleFactor' number of rows
+           int len = queryArray.length;
+           
+           for (int i = 0; i < len - bundleFactor + 1; i += bundleFactor) {
+               HashMap[] rowBundle = Arrays.copyOfRange(queryArray, i, i + bundleFactor);
```
fhcarter: Seems like it'd be more straightforward to say something like
```
int rowsRemaining = queryArray.length
while (rowsRemaining > 0) {
    int rowsToCopy = (bundleFactor <= rowsRemaining ? rowsRemaining : bundleFactor)
    ... Arrays.copyOf(queryArray, I, rowsToCopy)
    ...
    rowsRemaining -= rowsToCopy
    if (rowsRemaining <= 0) {
        // send 200
    } else {
       // send 100
    }
...
}
```
Then you can get rid of the "check for remainder thing"

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -277,14 +279,45 @@ public synchronized void executePolling(String pollQuery) {
...
+               }
+               lastRowBundle = rowBundle;
+           }
+           
+           // Check for remainder and send if necessary
+           if (len % bundleFactor != 0) {
+               HashMap[] rowBundle = Arrays.copyOfRange(queryArray, len - len % bundleFactor, len);
+               client.sendQueryResponse(200, replyAddress, rowBundle);
+               lastRowBundle = rowBundle;
+           }
```
fhcarter: can lose this with simpler array structure

jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBC.java
```diff
@@ -540,6 +555,95 @@ public void testDBReconnect() throws VantiqSQLException {
...
+        responseBody = (JsonArray) response.getBody();
+        assert responseBody.size() == 2000;
+        assert core.lastRowBundle.length == 500;
+        
+        // Query with bundleFactor that doesn't divide evenly into 2000 rows
+        params.put("bundleFactor", 600);
+        response = vantiq.query(testSourceName, params);
+        responseBody = (JsonArray) response.getBody();
+        assert responseBody.size() == 2000;
+        assert core.lastRowBundle.length == 200;
```
fhcarter: can probably do the remainders (both) as 
```
core.lastRowBundle.length = ( rowCount % bundleSize)
```
where rowCount is a variable set to, in this case 2000, and bundleSize is a variable set to 500.

Indeed, both of these can be in a helper that has those as parameters.  Since only 2, not really necessary, but to consider in the future.

jdbcSource/README.md
```diff
@@ -149,8 +149,14 @@ will be sent as a unique Notification.
...
-only be **SELECT STATEMENTS**. The data will be returned as an array of the resulting rows. The following is an 
-example of a Procedure created in VANTIQ Modelo querying against a JDBC Source.
+only be **SELECT STATEMENTS**. The data will be returned as an array of the resulting rows. A query parameter named 
+`bundleFactor` can be used to specify how many rows of data a given message will contain. This parameter is useful when 
+querying a very large number of rows, where the data is likely to exceed the maximum message size allowed by the VANTIQ 
+System. Specifying this parameter as part of the WITH clause allows the connector to stream the data using multiple messages, 
+each containing `bundleFactor` number of rows. If no `bundleFactor` is specified, then the default value of 1000 will be used. 
+If a `bundleFactor` of 0 is specified, then the connector will attempt to send all of the queried rows in a single message, 
+(**NOTE**: this can likely cause errors if the combined data exceeds the maximum message size). The following is an example of 
+a Procedure created in VANTIQ Modelo querying against a JDBC Source.
```
fhcarter: Too geeky :-)

Maybe

only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message contains some number of rows. 

A query parameter  named `bundleFactor` determines how many rows are bundled into each message.
Generally, the default value of 1000 will be fine.  However, in cases where rows may be very large, a smaller
value may be required.

The `bundleFactor` parameter is specified in the `WITH` clause, allowing each query to adjust as appropriate.
If the parameter is not specified, the default value (1000) is used -- that will be fine for most cases.  If a value of `0` is provided, then all rows will be placed into a single message (the use of this is discouraged as it may cause problems).  If the value is positive (and non-zero), then that many rows will be sent at a time.

From the perspective of consuming the rows, there is no visible difference here.  The `bundleFactor` parameter is present to allow control when returning very large rows.

namirfa:
@fhcarter 

Added fixes based on reviewer comments.

Issue #167, opened 2019-05-28
JDBC Source: Add enhancement to improve performance of publish/query handlers


namirfa:
Currently, the JDBC Connector handles publish/query requests serially. This enhancement would allow the respective handlers to run a pool of threads which actually execute the JDBC operations. This would allow for a higher throughput of publish/query requests being processed.

The connector's performance will still be limited by whatever is feasible for the DB, the machine running the connector, etc..

PR #168, opened 2019-06-05
Naf jdbc performance #167

namirfa:
Closes Issue #167 

Adding ability to process query/publish requests asynchronously. Added tests to ensure behavior is correct.

Note: There is no need to do something of the "diagnoseConnection" sort for the connection pool, this is all handled internally by HikariCP.

### DIFF
--- a/jdbcSource/README.md
+++ b/jdbcSource/README.md
@@ -89,8 +89,11 @@ The Configuration document may look similar to the following example:
           "general": {
              "username": "sqlUsername",
              "password": "sqlPassword",
-             "dbURL": "jdbc:mysql://localhost/myDB?useSSL=false&serverTimezone=UTC"
-             "pollTime": 3000
+             "dbURL": "jdbc:mysql://localhost/myDB?useSSL=false&serverTimezone=UTC",
+             "asynchronousProcessing": true,
+             "maxRunningThreads": 10,
+             "maxQueuedTasks" 20,
+             "pollTime": 3000,
              "pollQuery": "SELECT * FROM myTable"
           }
        }
@@ -104,6 +107,12 @@ The Configuration document may look similar to the following example:
 *   **username**: Required. This is the username that will be used to connect to the SQL Database.
 *   **password**: Required. This is the password that will be used to connect to the SQL Database.
 *   **dbURL**: Required. This is the URL corresponding to the SQL Database that you will connect to.
+*   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
+opposed to the default behavior which is synchronous.
+*   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads 
+running at any given point for query or publish requests, respectively. Must be a non-zero integer.
+*   **maxQueuedTasks**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of queued 
+tasks at any given point for query or publish requests, respectively. Must be a non-zero integer.
 *   **pollTime**: Optional. If specified, you must specify the pollQuery as well. This option allows you to specify a polling 
     rate indicating the frequency (in milliseconds) at which the pollQuery will be executed. The value must be a positive
     number greater than 0, (*i.e.* 3000 --> executing every 3 seconds).
@@ -261,14 +270,13 @@ SQLException, and contains the Error Message, SQL State, and Error Code from the
 In order to properly run the tests, you must create an environment variable named **JDBC\_DRIVER\_LOC** which points to the 
 appropriate JDBC Driver .jar file.
 
- Additionally, you must add properties to your _gradle.properties_ file 
-in the _~/.gradle_ directory. 
-These properties include the JDBC Database username, password, and URL.
-You must also add the Target VANTIQ Server URL, as well as an 
-Authentication Token for that server. The Target VANTIQ Server and Auth Token will be used to create a temporary VANTIQ 
-Source and Type, named _testSourceName_ and _testTypeName_ respectively. These names can optionally be configured by adding 
-`EntConTestSourceName` and `EntConTestTypeName` to the gradle.properties file. The following shows what the gradle.properties file 
-should look like:
+Additionally, you must add properties to your _gradle.properties_ file in the _~/.gradle_ directory. These properties include 
+the JDBC Database username, password, and URL. You must also add the Target VANTIQ Server URL, as well as an Authentication 
+Token for that server. The Target VANTIQ Server and Auth Token will be used to create a temporary VANTIQ Source, VANTIQ Type, 
+VANTIQ Topic, VANTIQ Procedure and VANTIQ Rule. They will be named _testSourceName_, _testTypeName_, _testTopicName_, 
+_testProcedureName_ and _testRuleName_ respectively. These names can optionally be configured by adding 
+`EntConTestSourceName`, `EntConTestTypeName`, `EntConTestTopicName`, `EntConTestProcedureName` and `EntConTestRuleName` to the 
+gradle.properties file. The following shows what the gradle.properties file should look like:
 
 ```
     EntConJDBCUsername=<yourUsername>
@@ -278,6 +286,9 @@ should look like:
     TestAuthToken=<yourAuthToken>
     EntConTestSourceName=<yourDesiredSourceName>
     EntConTestTypeName=<yourDesiredTypeName>
+    EntConTestTopicName=<yourDesiredTopicName>
+    EntConTestProcedureName=<yourDesiredProcedureName>
+    EntConTestRuleName=<yourDesiredRuleName>
 ```
 
 * **NOTE:** We strongly encourage users to create a unique VANTIQ Namespace in order to ensure that tests do not accidentally 
--- a/jdbcSource/README.md
+++ b/jdbcSource/README.md
@@ -110,9 +110,9 @@ The Configuration document may look similar to the following example:
 *   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
 opposed to the default behavior which is synchronous.
 *   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads 
-running at any given point for query or publish requests, respectively. Must be a non-zero integer.
+running at any given point for query or publish requests, respectively. Must be a non-zero integer. Default value is 5.
 *   **maxQueuedTasks**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of queued 
-tasks at any given point for query or publish requests, respectively. Must be a non-zero integer.
+tasks at any given point for query or publish requests, respectively. Must be a non-zero integer. Default value is 10.
 *   **pollTime**: Optional. If specified, you must specify the pollQuery as well. This option allows you to specify a polling 
     rate indicating the frequency (in milliseconds) at which the pollQuery will be executed. The value must be a positive
     number greater than 0, (*i.e.* 3000 --> executing every 3 seconds).
--- a/jdbcSource/README.md
+++ b/jdbcSource/README.md
@@ -91,7 +91,7 @@ The Configuration document may look similar to the following example:
              "password": "sqlPassword",
              "dbURL": "jdbc:mysql://localhost/myDB?useSSL=false&serverTimezone=UTC",
              "asynchronousProcessing": true,
-             "maxRunningThreads": 10,
+             "maxActiveTasks": 10,
              "maxQueuedTasks" 20,
              "pollTime": 3000,
              "pollQuery": "SELECT * FROM myTable"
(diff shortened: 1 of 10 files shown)


jdbcSource/README.md
```diff
@@ -104,6 +107,12 @@ The Configuration document may look similar to the following example:
 *   **username**: Required. This is the username that will be used to connect to the SQL Database.
 *   **password**: Required. This is the password that will be used to connect to the SQL Database.
 *   **dbURL**: Required. This is the URL corresponding to the SQL Database that you will connect to.
+*   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
+opposed to the default behavior which is synchronous.
+*   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads 
+running at any given point for query or publish requests, respectively. Must be a non-zero integer.
+*   **maxQueuedTasks**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of queued 
+tasks at any given point for query or publish requests, respectively. Must be a non-zero integer.
```
fhcarter: must be a positive integer

jdbcSource/README.md
```diff
@@ -104,6 +107,12 @@ The Configuration document may look similar to the following example:
 *   **username**: Required. This is the username that will be used to connect to the SQL Database.
 *   **password**: Required. This is the password that will be used to connect to the SQL Database.
 *   **dbURL**: Required. This is the URL corresponding to the SQL Database that you will connect to.
+*   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
+opposed to the default behavior which is synchronous.
+*   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads 
+running at any given point for query or publish requests, respectively. Must be a non-zero integer.
```
fhcarter: Must be a positive integer

(that is, I can't set `maxRunninigThreads` to -3)

jdbcSource/build.gradle
```diff
@@ -80,4 +80,8 @@ dependencies {
     // Use JUnit test framework
     testCompile 'junit:junit:4.12'
     testCompile project(path:":extjsdk", configuration:"testArtifacts")
+
+    // Used to create a connection pool if asynchronous processing has been specified for publish/query handlers
+    compile group: 'com.zaxxer', name: 'HikariCP', version: '3.3.1'
```
fhcarter: Need to add license stuff (license file, list it, etc.)

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -38,27 +44,49 @@
     
     // Timeout (in seconds) used to check if connection is still valid
     private static final int CHECK_CONNECTION_TIMEOUT = 5;
-    
+
+    // Timeout (in milliseconds) specifying how long ds.getConnection() will wait for a connection before timing out
+    private static final int CONNECTION_POOL_TIMEOUT = 5000;
+
+    // Used if asynchronous publish/query handling has been specified
+    private HikariConfig config = new HikariConfig();
```
fhcarter: We have lots of `config`s.  Might be good to give this a more descriptive name.

Also, need we allocate it if we aren't going to use it?

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -73,17 +101,30 @@ public void setupJDBC(String dbURL, String username, String password) throws Van
...
+        if (isAsync) {
+            try (Connection conn = ds.getConnection();
+                 Statement stmt = conn.createStatement();
+                 ResultSet rs = stmt.executeQuery(sqlQuery)) {
+                rsArray = createMapFromResults(rs);
+            } catch (SQLException e) {
+                // Handle errors for JDBC
+                reportSQLError(e);
+            }
+        } else {
```
fhcarter: Would it be better to just always use the connection pool?  Something like set it to a _pool of 1_ (`maximumPoolSize=1`)if we aren't async? Would that simplify the code?

Or do we want to do that only after this has been used for a while?

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -38,27 +44,49 @@
...
+            if (asyncProcessing) {
+                // Create a connection pool
+                config.setJdbcUrl(dbURL);
+                if (username != null) {
+                    config.setUsername(username);
+                }
+                if (password != null) {
+                    config.setPassword(password);
+                }
+                ds = new HikariDataSource(config);
```
fhcarter: Do we need to set the `maximumPoolSize` here?  Or allow user to override?

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
```diff
@@ -173,7 +226,7 @@ public int processPublish(String sqlQuery) throws VantiqSQLException {
     }
     
     /**
-     * Method used to try and reconnect if database connection was lost.
+     * Method used to try and reconnect if database connection was lost. Used for synchronous processing.
```
fhcarter: Might want to note that the connection pool does this on its own...

jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBC.java
```diff
@@ -649,9 +665,74 @@ public void testMaxMessageSize() throws VantiqSQLException {
...
+
+        // Check that Source, Type, Topic, Procedure and Rule do not already exist in namespace, and skip test if they do
+        assumeFalse(checkSourceExists());
+        assumeFalse(checkTypeExists());
+        assumeFalse(checkTopicExists());
+        assumeFalse(checkProcedureExists());
+        assumeFalse(checkRuleExists());
+
+        // Setup a VANTIQ JDBC Source, and start running the core
+        setupSource(createSourceDef(true));
```
fhcarter: should we also have a test when the optional parameters (max...Threads or whatever they're called) aren't provided?

jdbcSource/README.md
```diff
@@ -104,6 +107,12 @@ The Configuration document may look similar to the following example:
 *   **username**: Required. This is the username that will be used to connect to the SQL Database.
 *   **password**: Required. This is the password that will be used to connect to the SQL Database.
 *   **dbURL**: Required. This is the URL corresponding to the SQL Database that you will connect to.
+*   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
+opposed to the default behavior which is synchronous.
+*   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads
```
fhcarter: Would a better name be something like `maxActiveTasks`?  Then it's parallel to `maxQueuedTasks` and can describe what's actually happening.  "I can queue up some things, but only n of them can be active at the same time"

jdbcSource/README.md
```diff
@@ -104,6 +107,12 @@ The Configuration document may look similar to the following example:
 *   **username**: Required. This is the username that will be used to connect to the SQL Database.
 *   **password**: Required. This is the password that will be used to connect to the SQL Database.
 *   **dbURL**: Required. This is the URL corresponding to the SQL Database that you will connect to.
+*   **asynchronousProcessing**: Optional. If set to `true`, query and publish requests will be handled asynchronously, as 
+opposed to the default behavior which is synchronous.
+*   **maxRunningThreads**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of threads 
+running at any given point for query or publish requests, respectively. Must be a non-zero integer. Default value is 5.
+*   **maxQueuedTasks**: Optional. Only used if `asynchronousProcessing` is set to `true`. The maximum number of queued
```
fhcarter: Both of these must be a _positive_ integer. 

Separate:  are queue tasks running?  That is, is there a relationship between the two numbers?  Can `maxQueuedTasks` be zero, meaning if there are no runners, throw an error?

(Not a request to provide that, just want to know)

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCHandleConfiguration.java
```diff
@@ -214,6 +208,107 @@ public void run() {
...
+        if (generalConfig.get(ASYNCH_PROCESSING) instanceof Boolean && (Boolean) generalConfig.get(ASYNCH_PROCESSING)) {
+            asynchronousProcessing = true;
+            int maxActiveTasks = MAX_ACTIVE_TASKS;
+            int maxQueuedTasks = MAX_QUEUED_TASKS;
+
+            if (generalConfig.get(MAX_ACTIVE) instanceof Integer && (Integer) generalConfig.get(MAX_ACTIVE) > 0) {
+                maxActiveTasks = (Integer) generalConfig.get(MAX_ACTIVE);
+
+                // Used to set the max pool size for connection pool
+                maxPoolSize = 2*maxActiveTasks;
```
fhcarter: This sets `maxPoolSize` only if the user provides a `maxActiveTasks` setting.  Seems like we should have a default here?  Probably the same `2 * maxActiveTasks` is fine, but should be set regardless.  Zero's applicable only in the non-async case, I think.  Zero is probably never a valid pool size value.

(This is why we need the tests for taking the defaults in the configuration -- if not provided, it'll either fail or use some default that's not what we expect.)

jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBCConfig.java
```diff
@@ -122,6 +122,41 @@ public void testPollingConfig() {
...
+        // Setting asynchronousProcessing to true
+        conf.put("asynchronousProcessing", true);
+        sendConfig(conf, vantiqConf);
+        assertFalse("Should not fail with asynchronousProcessing set to true", configIsFailed());
+
+        // Setting maxRunningThreads and maxQueuedTasks incorrectly
+        conf.put("maxActiveTasks", "jibberish");
+        conf.put("maxQueuedTasks", "moreJibberish");
+        sendConfig(conf, vantiqConf);
+        assertFalse("Should not fail when maxRunningThreads and maxQueuedTasks are set incorrectly", configIsFailed());
```
fhcarter: minor -- name change in error text

jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBCConfig.java
```diff
@@ -122,6 +122,41 @@ public void testPollingConfig() {
...
+
+        // Setting maxRunningThreads and maxQueuedTasks correctly
+        conf.put("maxActiveTasks", 10);
+        conf.put("maxQueuedTasks", 20);
+        sendConfig(conf, vantiqConf);
+        assertFalse("Should not fail when maxRunningThreads and maxQueuedTasks are set correctly", configIsFailed());
```
fhcarter: ditto

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCHandleConfiguration.java
```diff
@@ -214,6 +208,107 @@ public void run() {
...
+
+            if (generalConfig.get(MAX_QUEUED) instanceof Integer && (Integer) generalConfig.get(MAX_QUEUED) > 0) {
+                maxQueuedTasks = (Integer) generalConfig.get(MAX_QUEUED);
+            }
+
+            // Creating the thread pool executors with Queue
+            source.queryPool = new ThreadPoolExecutor(maxActiveTasks, maxActiveTasks, 0l, TimeUnit.MILLISECONDS,
+                    new LinkedBlockingQueue<Runnable>(maxQueuedTasks));
+            source.publishPool = new ThreadPoolExecutor(maxActiveTasks, maxActiveTasks, 0l, TimeUnit.MILLISECONDS,
+                    new LinkedBlockingQueue<Runnable>(maxQueuedTasks));
```
fhcarter: We can add the no `LinkedBlockingQueue` case when `maxQueuedTasks` is zero later.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCHandleConfiguration.java
namirfa: The reason I only set the value here is because my intention was to match whatever `2 * maxActiveTasks` would be. In this case, we make sure that we will have enough connections in our connection pool to match exactly twice the number of `maxActiveTasks` that was specified in the Source Configuration. Otherwise, our default value for `maxActiveTasks` is 5, and the default max of connections in the connection pool is 10 (so we don't have to set it in `JDBC.setupJDBC()`).

Let me know if that makes sense. I should add a comment inline to explain this behavior.

In any case, I am adding tests for the defaults now.

fhcarter:
I would make it explicit. One of the defaults could change. Accidental matches are fragile

Sent from my iPhone

On Jun 6, 2019, at 10:06 AM, namirfa <notifications@github.com<mailto:notifications@github.com>> wrote:


@namirfa commented on this pull request.

________________________________

In jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCHandleConfiguration.java<https://github.com/Vantiq/vantiq-extension-sources/pull/168#discussion_r291282654>:

> +     *                          If default active tasks is used, then returns 0.
+     */
+    private int createQueryAndPublishHandlers(Map<String, ?> generalConfig) {
+        int maxPoolSize = 0;
+
+        // Checking if asynchronous processing was specified in the general configuration
+        if (generalConfig.get(ASYNCH_PROCESSING) instanceof Boolean && (Boolean) generalConfig.get(ASYNCH_PROCESSING)) {
+            asynchronousProcessing = true;
+            int maxActiveTasks = MAX_ACTIVE_TASKS;
+            int maxQueuedTasks = MAX_QUEUED_TASKS;
+
+            if (generalConfig.get(MAX_ACTIVE) instanceof Integer && (Integer) generalConfig.get(MAX_ACTIVE) > 0) {
+                maxActiveTasks = (Integer) generalConfig.get(MAX_ACTIVE);
+
+                // Used to set the max pool size for connection pool
+                maxPoolSize = 2*maxActiveTasks;


The reason I only set the value here is because my intention was to match whatever 2 * maxActiveTasks would be. In this case, we make sure that we will have enough connections in our connection pool to match exactly twice the number of maxActiveTasks that was specified in the Source Configuration. Otherwise, our default value for maxActiveTasks is 5, and the default max of connections in the connection pool is 10 (so we don't have to set it in JDBC.setupJDBC()).

Let me know if that makes sense. I should add a comment inline to explain this behavior.

In any case, I am adding tests for the defaults now.

namirfa:
@fhcarter 
Added fixes for all of the reviewer comments. Both asynchronous processing tests pass as expected.

fhcarter:
I'll look tonight. Hard to review on a phone!

namirfa:
No problem, thanks!

Issue #169, opened 2019-06-06
JDBC Source: Batch Inserts


namirfa:
Add an enhancement to the JDBC Source in order to handle batch inserts. Currently, INSERTS are expected to be represented as a single string sent to the source via a publish request. This enhancement would allow the source to accept an array of strings, which would then be processed as a batch insert.

PR #170, opened 2019-06-10
Naf jdbc batch #169

namirfa:
Closes Issue #169 

An enhancement that enables batch inserts.

If a publish request provides the query as a list of strings, then the list will be processed as a batch. Otherwise, all previous behavior is identical. Each database has a different value for the max batch size, so I don't think we can try to enforce any maximum from our end. Instead, if the max size is reached then we will throw a VantiqSQLException and handle it normally.

### DIFF
--- a/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
+++ b/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
@@ -9,6 +9,7 @@
 package io.vantiq.extsrc.jdbcSource;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -220,8 +221,8 @@ public class JDBCCore {
     }
     
     /**
-     * Executes the query that is provided as a String in the options specified by the "query" key in the
-     * object of the Publish message.
+     * Executes the query that is provided in the Publish Message. If query is an Array of Strings, then it is executed as a Batch request.
+     * If the query is a single String, then it is executed normally.
      * @param message   The Query message.
      */
     public void executePublish(ExtensionServiceMessage message) {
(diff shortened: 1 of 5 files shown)


jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -243,12 +244,19 @@ public void executePublish(ExtensionServiceMessage message) {
                 String queryString = (String) request.get("query");
                 int data = localJDBC.processPublish(queryString);
                 log.trace("The returned integer value from Publish Query is the following: ", data);
+            } else if (request.get("query") instanceof List) {
```
fhcarter: can you check here for `instanceof List<String>` -- since a list of integers as it stands isn't gonna work well.

namirfa: I wanted to do this, but Java only lets me check for a generic list. I have a test that handles a list of integers, and I added a catch block for "ClassCastException" so at least we can have a better error message.

Let me know if that's alright.

fhcarter:
That's what there is. So that's fine

Sent from my iPhone

On Jun 11, 2019, at 12:36 PM, namirfa <notifications@github.com<mailto:notifications@github.com>> wrote:


@namirfa commented on this pull request.

________________________________

In jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java<https://github.com/Vantiq/vantiq-extension-sources/pull/170#discussion_r292552065>:

> @@ -243,12 +244,19 @@ public void executePublish(ExtensionServiceMessage message) {
                 String queryString = (String) request.get("query");
                 int data = localJDBC.processPublish(queryString);
                 log.trace("The returned integer value from Publish Query is the following: ", data);
+            } else if (request.get("query") instanceof List) {


I wanted to do this, but Java only lets me check for a generic list. I have a test that handles a list of integers, and I added a catch block for "ClassCastException" so at least we can have a better error message.

Let me know if that's alright.

PR #174, opened 2019-07-09
Naf obj opencv upgrade #110

### LINKED ISSUE #110: Add support for OpenCV 4.0.1 for object recognition extension source
Add support for OpenCV 4.0.1 for object recognition extension source

namirfa:
Upgrading both the tensorflow and opencv versions to the most recent ones. Closes #110.

@fhcarter The project builds as expected, and all tests pass. On occasion, one or two random tests will fail when building, but all tests pass when run explicitly.

### DIFF
--- a/objectRecognitionSource/build.gradle
+++ b/objectRecognitionSource/build.gradle
@@ -50,7 +50,7 @@ def standardImplementationTestPatterns = new ArrayList<String>(
 
 // This is the version of OpenCV that you will be using. Should be "XYZ" for version X.Y.Z
 
-def opencvVersion = "342"
+def opencvVersion = "410"
 
 if (System.env.OPENCV_LOC != null ) {
     def ocv = new File(System.env.OPENCV_LOC)
@@ -197,7 +197,7 @@ dependencies {
     
     // Used by YoloProcessor alone. These two dependencies and all files in the edu.* package may be removed if YoloProcessor
     // is removed
-    compile "org.tensorflow:tensorflow:1.9.0"
+    compile "org.tensorflow:tensorflow:1.13.1"
     compile "org.apache.commons:commons-math3:3.6.1"
     
     // Used by FtpRetriever alone. Can be removed if FtpRetriever is removed
--- a/objectRecognitionSource/build.gradle
+++ b/objectRecognitionSource/build.gradle
@@ -197,7 +197,7 @@ dependencies {
     
     // Used by YoloProcessor alone. These two dependencies and all files in the edu.* package may be removed if YoloProcessor
     // is removed
-    compile "org.tensorflow:tensorflow:1.13.1"
+    compile "org.tensorflow:tensorflow:1.11.0"
     compile "org.apache.commons:commons-math3:3.6.1"
     
     // Used by FtpRetriever alone. Can be removed if FtpRetriever is removed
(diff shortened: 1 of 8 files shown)


objectRecognitionSource/build.gradle
```diff
@@ -197,7 +197,7 @@ dependencies {
     
     // Used by YoloProcessor alone. These two dependencies and all files in the edu.* package may be removed if YoloProcessor
     // is removed
-    compile "org.tensorflow:tensorflow:1.9.0"
+    compile "org.tensorflow:tensorflow:1.13.1"
```
fhcarter: Before we check this in, we should check with Brett and/or Patrick.  TensorFlow versions >= 1.12 require a GPU with a CUDA Count (a measure of the GPU's power) greater than 6.  Some of the machines they use may not have that power (those are represented in "Data Center" class GPUs, with costs much higher than typical PC GPU prices).  So we might want to think carefully here.

1.11.* is still OK with the GPUs of mere mortals

namirfa:
@fhcarter 
I changed the tensorflow version to 1.11.0

I have tried to replicate the occasional test failures, but haven't had any luck. Not sure why all of a sudden the build is successful every time.

namirfa:
@fhcarter 
Fixed tests to avoid the "random" failures that were related to timing with the VANTIQ SDK. I have run the build about 5 times consecutively (takes ~10 min), and it succeed every time.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestNoProcessorQueries.java
```diff
@@ -311,14 +313,25 @@ public static void deleteSource() {
...
+                if (++retries < maxRetries) {
+                    done = false;
+                    Thread.sleep(50);
+                } else {
+                    List<VantiqError> errors = vantiqResponse.getErrors();
+                    for (int i = 0; i < errors.size(); i++) {
+                        if (errors.get(i).getCode().equals(NOT_FOUND_CODE)) {
+                            fail("Image should have been uploaded to VANTIQ");
+                        }
+                    }
```
fhcarter: Seems like we should be able to stuff these in a helper class or superclass so that we need only one copy of them.

They should go into `NeuralNetBase` so that they're available everywhere.  If they need parameterization (pass in `vantiq` or whatever, or move that to `NeuralNetBase`, too), that's OK.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestUploadAndDeleteQuery.java
```diff
@@ -682,22 +684,44 @@ public void testImageDateDeleteRange() {
...
+            done = true;
+            vantiqResponse = vantiq.selectOne("system.documents", name);
+            if (vantiqResponse.isSuccess()) {
+                if (++retries < maxRetries) {
+                    done = false;
+                    Thread.sleep(50);
+                } else {
+                    fail("Image should not have been uploaded to VANTIQ");
+                }
+            }
```
fhcarter: Move dups to super

namirfa:
@fhcarter 
Removed the duplicate code as you mentioned. I ended up just passing the vantiq instance as a parameter to the method, since I had already done well over half of the changes before you stopped by.

PR #175, opened 2019-07-11
Naf jdbc override query #171

### LINKED ISSUE #171: JDBC: Return errors from publish requests
Currently, all errors resulting from publish requests are logged locally, but not returned to VANTIQ. This enhancement would enable the JDBC Source to return those errors to VANTIQ so that they can be handled appropriately, without having to check the local log file.

namirfa:
Closes #171 

Adds an enhancement which allows Vail Select Statements to update the database, in addition to querying the database. This is accomplished by adding the `isUpdate` parameter to the select statements.

All tests pass as expected.

### DIFF
--- a/jdbcSource/README.md
+++ b/jdbcSource/README.md
@@ -157,9 +157,14 @@ will be sent as a unique Notification.
 ## Select Statements
 
 In order to interact with the JDBC Source, one option is to use VAIL to select from the source. To do this, you will need 
-to specify the SQL Query you wish to execute against your database as part of the WITH clause. The SQL Queries used here must 
-only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message contains some 
-number of rows.
+to specify the SQL Query you wish to execute against your database as part of the WITH clause. *Typically*, the SQL Queries 
+used here must only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message 
+contains some number of rows.
+
+However, if the `isUpdate` parameter is specified and set to `true`, the SQL Queries can be any valid SQL Update Statement 
+(i.e. `CREATE`, `INSERT`, `DELETE`, `DROP`, etc...). This feature mimics the behavior of the Publish Statements, which are 
+defined [below](#publish). In this case, the VAIL Select Statements will return an empty map if the statement was executed 
+successfully, or a query error otherwise.
 
 A query parameter named `bundleFactor` determines how many rows are bundled into each message. Generally, the default value of 
 500 will be fine. However, in cases where rows may be very large, a smaller value may be required.
@@ -172,6 +177,7 @@ rows will be placed into a single message (the use of this is discouraged as it 
 From the perspective of consuming the rows, there is no visible difference here. The `bundleFactor` parameter is present to 
 allow control when returning very large rows.
 
+The following example uses a Vail Select Statement to **query** a database:
 ```
 PROCEDURE queryJDBC()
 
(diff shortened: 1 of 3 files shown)


jdbcSource/README.md
```diff
@@ -157,9 +157,14 @@ will be sent as a unique Notification.
 ## Select Statements
 
 In order to interact with the JDBC Source, one option is to use VAIL to select from the source. To do this, you will need 
-to specify the SQL Query you wish to execute against your database as part of the WITH clause. The SQL Queries used here must 
-only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message contains some 
-number of rows.
+to specify the SQL Query you wish to execute against your database as part of the WITH clause. *Typically*, the SQL Queries 
+used here must only be **SELECT STATEMENTS**. The data will be returned to VANTIQ as a set of messages, where each message
```
fhcarter: "must only" --> "would"

"typically ... must" doesn't make sense.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -192,17 +192,38 @@ public void executeQuery(ExtensionServiceMessage message) {
...
             }
         }
-        
+
+        // Check if query request is actually an update
+        boolean isUpdate = false;
+        if (request.get("isUpdate") instanceof Boolean && (Boolean) request.get("isUpdate")) {
+            isUpdate = (Boolean) request.get("isUpdate");
+        }
+
```
fhcarter: Just set `isUpdate` here by querying the query?  Then no other code must change?

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -192,17 +192,38 @@ public void executeQuery(ExtensionServiceMessage message) {
...
+                    // Send empty response back
+                    client.sendQueryResponse(204, replyAddress, new LinkedHashMap<>());
+                } else {
+                    HashMap[] queryArray = localJDBC.processQuery(queryString);
+                    sendDataFromQuery(queryArray, message);
+                }
+            } else if (isUpdate && request.get("query") instanceof List) {
+                List queryArray = (List) request.get("query");
+                int[] data = localJDBC.processBatchPublish(queryArray);
+                log.trace("The returned integer array from Publish Query is the following: ", data);
```
fhcarter: In batch case, aren't things always updates?  Do we do batch selects?

namirfa:
@fhcarter 

Added fixes based on your comments. All the tests pass and the build works successfully.

jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBCCore.java
```diff
@@ -192,17 +193,42 @@ public void executeQuery(ExtensionServiceMessage message) {
...
-        
+
         // Gather query results and send the appropriate response, or send a query error if an exception is caught
         try {
             if (request.get("query") instanceof String) {
                 String queryString = (String) request.get("query");
-                HashMap[] queryArray = localJDBC.processQuery(queryString);
-                sendDataFromQuery(queryArray, message);
+                // Check if SQL Query is an update statement, or query statement
+                if (!queryString.trim().toLowerCase().startsWith(SELECT_STATEMENT_IDENTIFIER)) {
```
rblumer4: A style/readability tip:
if you have both IF and and ELSE clauses, I think it's more readable if you don't put the NOT in the test.
E.g.  I would reverse this IF and its clauses:  
     if (queryString.trim().toLowerCase().startsWith(SELECT_STATEMENT_IDENTIFIER)) ...

Other things I try to do:  
-I try to put the "common case" into the IF clause and the "exception case" into the ELSE clause...
-When one clause has 2-3 lines and the other has more lines, I try to put the smaller/simpler clause in  the IF half, so you can quickly see that & then move on to reading the ELSE clause.

namirfa: Thanks for the advice Roger! I'll fix this later this afternoon.

PR #176, opened 2019-07-11
Naf obj polling fix cleanup

### LINKED ISSUE #177: Object Recognition Source: Add Parallel Processing for Polling
Currently, (when the Object Recognition Source is configured to poll an image source), it captures a frame from the image source, and then synchronously processes the frame before returning the results. This process is repeated for each captured frame.

This enhancement allows the image processing to be done asynchronously, using a pool of threads. As a result, the source can now capture frames from the image source, and then process/return data from multiple different frames in parallel. This improves the performance of the source.

The user has the ability to configure the `maxRunningThreads`, as well as the `maxQueuedTasks`. The default values, respectively, are 10 and 20. If the number of `maxQueuedTasks` is exceeded, the source will drop the oldest unexecuted request and replace it with the newer one.

namirfa:
Closes #177 

@fhcarter 

In order to clean up the mess I made with the last PR, I moved all of the relevant commits into this new branch. This PR contains all of the work that had been done in the previous PR, and additionally adds some extra debugging/error handling, as well as new tests to ensure the DiscardOldestPolicy works as intended.

All tests pass and build is successful.

### DIFF
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -188,9 +188,11 @@ the previous is sent.
     will be valid General Config options, but in the future only "pollTime" will be supported.)
 *   allowQueries: This option allows Queries to be received when set to `true'
 *   maxRunningThreads: Optional. Only used if `pollTime` has been specified. The maximum number of threads running at any 
-given point for polling requests. Must be a positive integer. Default value is 10.
+given point for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 10.
 *   maxQueuedTask: Optional. Only used if `pollTime` has been specified. The maximum number of queued tasks at any given point 
-for polling requests. Must be a positive integer. Default value is 20.
+for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 20.
+    *   **NOTE:** The default behavior of the Object Recognition Source is to process the captured frames in parallel. If you 
+    would like the source to process frames sequentially, then `maxRunningThreads` and `maxQueuedTasks` must both be set to 1.
 
 ### Options Available for Data Source
 
--- a/objectRecognitionSource/README.md
+++ b/objectRecognitionSource/README.md
@@ -191,8 +191,9 @@ the previous is sent.
 given point for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 10.
 *   maxQueuedTask: Optional. Only used if `pollTime` has been specified. The maximum number of queued tasks at any given point 
 for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 20.
-    *   **NOTE:** The default behavior of the Object Recognition Source is to process the captured frames in parallel. If you 
-    would like the source to process frames sequentially, then `maxRunningThreads` and `maxQueuedTasks` must both be set to 1.
+    *   **NOTE:** The default behavior of the Object Recognition Source is to process up to 10 captured frames in parallel at 
+    a time. If you would like the source to only process one frame at a time, then `maxRunningThreads` and `maxQueuedTasks` 
+    must both be set to 1.
 
 ### Options Available for Data Source
 
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -52,6 +52,7 @@ public class ObjectDetector {
     private int saveRate = 0;
     private Boolean labelImage;
     private int frameCount = 0;
+    int fileCount = 0; // Used for saving files with same name
     private float threshold;
     private Vantiq vantiq = null;
     private String sourceName = null;
@@ -163,15 +164,20 @@ public class ObjectDetector {
      *              identification is accurate, and a {@code location} containing the coordinates for the
      *              {@code top},{@code left}, {@code bottom}, and {@code right} edges of the bounding box for the object.
      */
-    public List<Map<String, ?>> detect(final byte[] image) {
+    public List<Map<String, ?>> detect(final byte[] image, Date timestamp) {
         try (Tensor<Float> normalizedImage = normalizeImage(image)) {
-            Date now = new Date(); // Saves the time before
             List<Recognition> recognitions = YOLOClassifier.getInstance(threshold, anchorArray, frameSize).classifyImage(executeYOLOGraph(normalizedImage), labels);
             BufferedImage buffImage = imageUtil.createImageFromBytes(image);
             
             // Saves an image every saveRate frames
             if (imageUtil.saveImage && ++frameCount >= saveRate) {
-                String fileName = format.format(now) + ".jpg";
+                String fileName = format.format(timestamp);
+                if (lastFilename != null && lastFilename.contains(fileName)) {
+                    fileName = fileName + "(" + ++fileCount + ").jpg";
+                } else {
+                    fileName = fileName + ".jpg";
+                    fileCount = 0;
+                }
                 lastFilename = fileName;
                 if (labelImage) {
                     buffImage = imageUtil.labelImage(buffImage, recognitions);
@@ -211,17 +217,12 @@ public class ObjectDetector {
             BufferedImage buffImage = imageUtil.createImageFromBytes(image);
             
             // Saves an image if requested
-            if (outputDir != null || vantiq != null || (fileName != null && this.imageUtil.saveImage)) {
+            if (outputDir != null || vantiq != null || this.imageUtil.saveImage) {
                 ImageUtil imageUtil = new ImageUtil();
                 imageUtil.outputDir = outputDir;
                 imageUtil.vantiq = vantiq;
                 imageUtil.sourceName = sourceName;
                 imageUtil.frameSize = frameSize;
-                if (fileName == null) {
-                    fileName = format.format(now) + ".jpg";
-                } else if (!fileName.endsWith(".jpg") && !fileName.endsWith(".jpeg")) {
-                    fileName += ".jpg";
-                }
                 lastFilename = fileName;
                 if (labelImage) {
                     buffImage = imageUtil.labelImage(buffImage, recognitions);
--- a/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
+++ b/objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
@@ -158,11 +158,12 @@ public class ObjectDetector {
     /**
      * Detect objects on the given image
      * <br>Edited to return the results as a map and conditionally save the image
-     * @param image The image in jpeg format
-     * @return      A List of Maps, each of which has a {@code label} stating the type of the object identified,
-     *              a {@code confidence} specifying on a scale of 0-1 how confident the neural net is that the
-     *              identification is accurate, and a {@code location} containing the coordinates for the
-     *              {@code top},{@code left}, {@code bottom}, and {@code right} edges of the bounding box for the object.
+     * @param image     The image in jpeg format
+     * @param timestamp The timestamp corresponding to when the frame was captured. Used to name the image if it is being saved
+     * @return          A List of Maps, each of which has a {@code label} stating the type of the object identified,
+     *                  a {@code confidence} specifying on a scale of 0-1 how confident the neural net is that the
+     *                  identification is accurate, and a {@code location} containing the coordinates for the
+     *                  {@code top},{@code left}, {@code bottom}, and {@code right} edges of the bounding box for the object.
      */
     public List<Map<String, ?>> detect(final byte[] image, Date timestamp) {
         try (Tensor<Float> normalizedImage = normalizeImage(image)) {
--- a/objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
+++ b/objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
@@ -15,6 +15,7 @@ import java.util.Map;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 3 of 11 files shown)


namirfa:
Adding README changes now, to show how to configure the `maxRunningThreads` and `maxQueuedTasks`.

fhcarter:
There should be an issue for this & PR should say fixes....  And branch should be named appropriately.  Branch is OK, but please add the issue & reference it.  AS it is, not really sure what this PR is for.  PR doesn't say.

objectRecognitionSource/README.md
```diff
@@ -185,6 +187,10 @@ the previous is sent.
     *   (**NOTE:** Previously named "pollRate". For a limited amount of time, both "pollTime" and "pollRate"
     will be valid General Config options, but in the future only "pollTime" will be supported.)
 *   allowQueries: This option allows Queries to be received when set to `true'
+*   maxRunningThreads: Optional. Only used if `pollTime` has been specified. The maximum number of threads running at any 
+given point for polling requests. Must be a positive integer. Default value is 10.
```
fhcarter: Is this max threads/queued tasks per source?  per connector?

objectRecognitionSource/src/main/java/edu/ml/tensorflow/ObjectDetector.java
```diff
@@ -163,15 +164,20 @@ public ObjectDetector(float thresh, String graphFile, String labelFile, String m
      *              identification is accurate, and a {@code location} containing the coordinates for the
      *              {@code top},{@code left}, {@code bottom}, and {@code right} edges of the bounding box for the object.
      */
-    public List<Map<String, ?>> detect(final byte[] image) {
+    public List<Map<String, ?>> detect(final byte[] image, Date timestamp) {
```
fhcarter: new parameter missing in procedure header comment.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -250,6 +259,8 @@ boolean createNeuralNet(Map<String, ?> neuralNetConfig ) {
                 neuralNetType = YOLO_PROCESSOR_FQCN;
             } else if (neuralNetType.equals(NONE)) {
                 neuralNetType = NO_PROCESSOR_FQCN;
+            } else if (neuralNetType.equals(TEST)) {
+                neuralNetType = TEST_PROCESSOR_FQCN;
             } else if (neuralNetType.equals(DEFAULT)) {
                 neuralNetType = DEFAULT_NEURAL_NET;
```
fhcarter: I take it that users cannot add their own here -- they have to add a NeuralNetInterface?  Though we've never had anyone ask (& it seems unlikely in the near term), I thought we advertised the ability.  Maybe they have to go change this code.  But it seems like something we should be able to handle in configuration & dynamic loading.

Add an issue...

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -346,11 +357,30 @@ boolean createImageRetriever(Map<String, ?> dataSourceConfig) {
...
+        if (general.get("maxRunningThreads") instanceof Integer && (Integer) general.get("maxRunningThreads") > 0) {
+            maxRunningThreads = (Integer) general.get("maxRunningThreads");
+        }
+        
+        if (general.get("maxQueuedTasks") instanceof Integer && (Integer) general.get("maxQueuedTasks") > 0) {
+            maxQueuedTasks = (Integer) general.get("maxQueuedTasks");
+        }
+                
+        source.pool = new ThreadPoolExecutor(maxRunningThreads, maxRunningThreads, 0l, TimeUnit.MILLISECONDS, 
+                new LinkedBlockingQueue<Runnable>(maxQueuedTasks), new ThreadPoolExecutor.DiscardOldestPolicy());
```
fhcarter: Should we document that if you DON'T want the background execution, you should set the maxThreads & maxQueud to 1, 0, respectively?  But that will generate errors, so we really don't have a BW compat way.  Which is probably OK with our single user, but we have to start thinking about all this...

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
```diff
@@ -369,7 +399,7 @@ public void run() {
             source.pollTimer.schedule(task, 0, pollRate);
         } else if (polling == 0) {
             source.pollTimer = new Timer("dataCapture");
-            source.pollTimer.scheduleAtFixedRate(task, 0, 1);
+            source.pollTimer.schedule(task, 0, 1);
```
fhcarter: What are we doing here?  (Need a comment in the code rather than answering this...)

Is this doing constant/immediate polling?

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -79,7 +80,8 @@
     private static final String UPLOAD = "upload";
     private static final String DELETE = "delete";
     private static final String FILTER = "filter";
-    
+    final static String SYNCH_KEY = "synchKey";
+
```
fhcarter: since this is a static, this is single threading for ALL instances of ObjectRecognitionCore.  That's probably OK, but that's what it's doing.

Since it seems to be used to sync to get local copy of the neuralNet variable (which is an instance variable), making this static made the synchronization more draconian.

In truth, if I've read everything right, sync(this) is fine here.  Or make SYNC_KEY non-static, and change the name to NEURAL_NET_SYNC or something.  & Add a comment re: what it's doing.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -304,25 +306,35 @@ public void sendDataFromImage(ImageRetrieverResults imageResults) {
...
         }
         
+        Map<String, Object> processingParams = new LinkedHashMap<>();
+        processingParams.put("imageTimestamp", imageResults.getTimestamp());
+        
         // Send the results of the neural net if it doesn't error out
+        NeuralNetInterface localNeuralNet = null;
         try {
-            synchronized (this) {
+            synchronized (SYNCH_KEY) {
```
fhcarter: are we purposely changing from synchronizing only within this instance to across all instances?  The SYNC_KEY isn't really adding much.  The notion of the named string or object is useful if there are multiple independent entities within, say, a class that need to be independently synchronized.  In that case, sync on the class or instance will work, but it's overly confining.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
```diff
@@ -647,6 +670,10 @@ public void close() {
             pollTimer.cancel();
             pollTimer = null;
         }
+        if (pool != null) {
+            pool.shutdownNow();
+            pool = null;
+        }
```
fhcarter: (Not about this change but about close()) -- Doesn't close need to sync on whatever's being used to snarf the neuralNet variable?  Otherwise, everyone else syncing on it is not of much value.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public NeuralNetResults processImage(byte[] image, Map<String, ?> request) throws ImageProcessingException {
+        processImage(null, image, request);
+        return new NeuralNetResults();
+    }
+
+    @Override
+    public NeuralNetResults processImage(Map<String, ?> processingParams, byte[] image) throws ImageProcessingException {
+        try {
+            Thread.sleep(sleepTime);
```
fhcarter: what's the purpose of the sleeps?  We should document what's going on in this class.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestParallelProcessing.java
```diff
@@ -0,0 +1,220 @@
...
+        setupSource(createSourceDef(useCustomTaskConfig));
+
+        // Create Type to store results
+        setupType();
+
+        // Create Rule to store results in Type
+        setupRule();
+
+        // Wait for 10 seconds
+        Thread.sleep(10000);
```
fhcarter: probably add some comment that says something about the source acting independently so we're awaiting it's doing something?  I'm guessing.

objectRecognitionSource/README.md
namirfa: This is the max threads/queued tasks per source. Each source has its own unique thread pool.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
namirfa: As it stands, the users would have to add their own NeuralNetInterface implementation, and then edit this portion of the code in order to use it.

I will add an issue to make the second part easier, since they shouldn't have to edit this code in order to actually use their implementation.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
namirfa: I agree that we should document this. The user would have to set both of those values to 1, since setting either one to 0 would throw an exception according to the documentation.

This enhancement will change the default behavior of the Object Rec source  to always process images in parallel, unless otherwise specified.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
namirfa: I will add a comment in the code as well, but figured I'd explain here too. This change means that instead of scheduling all tasks at a fixed rate from the first task, we will schedule them relative to the previous execution.

This means that if there are any delays while executing a task, the next task will still be scheduled `pollTime` milliseconds after the previous task has completed.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
namirfa: Understood. I think the simplest option would be to just remove the SYNCH_KEY and  synchronize on (this) instead.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionCore.java
namirfa: Will add the synch now.

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestProcessor.java
namirfa: I'll add documentation in this class to explain what is going on. The idea was that to ensure the the DiscardOldestPolicy was working as intended, we could use this class to mimic processing an image, but we would know exactly how long the "image processing" would take. 

This helped with writing a test to make sure that the maxRunningThreads/QueuedTasks and then DiscardOldestPolicy were both working as expected.

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestParallelProcessing.java
namirfa: I'll add a comment in the code. Basically, we wait for 10 seconds while the source polls for images, and then we check the results to make sure that the maxRunningThreads/QueuedTasks work appropriately, and that the DiscardOldestPolicy works as well.

namirfa:
@fhcarter 
Added the fixes based on your comments.

objectRecognitionSource/README.md
```diff
@@ -185,6 +187,12 @@ the previous is sent.
     *   (**NOTE:** Previously named "pollRate". For a limited amount of time, both "pollTime" and "pollRate"
     will be valid General Config options, but in the future only "pollTime" will be supported.)
 *   allowQueries: This option allows Queries to be received when set to `true'
+*   maxRunningThreads: Optional. Only used if `pollTime` has been specified. The maximum number of threads running at any 
+given point for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 10.
+*   maxQueuedTask: Optional. Only used if `pollTime` has been specified. The maximum number of queued tasks at any given point 
+for polling requests from the specific VANTIQ source. Must be a positive integer. Default value is 20.
+    *   **NOTE:** The default behavior of the Object Recognition Source is to process the captured frames in parallel. If you 
+    would like the source to process frames sequentially, then `maxRunningThreads` and `maxQueuedTasks` must both be set to 1.
```
fhcarter: Have we tried this?  That is, I suspect what happens here is that we process one image.  If another comes along, it will fail to queue (if we're still processing) and add some error but drop that image.

Or, is what will happen that we continually replace the queued task until the first one finishes, at which point we'll process the "latest" queued task (due to discardOld policy)?

objectRecognitionSource/src/main/java/io/vantiq/extsrc/objectRecognition/ObjectRecognitionConfigHandler.java
fhcarter: I suspect (as noted above) that parallel (albeit limited) is now the only possibility.  That's probably fine, but we should verify that that's true & document.

objectRecognitionSource/README.md
namirfa: The behavior is the latter.

We will set up a queue with a size of one, and allow one thread to run at a time. Assuming the pollTime is faster than the time to execute a task, the queue will continually replace old tasks with the newest poll request until the original task as finished executing.

So in effect, the frames are being processed one at a time, but it is not true "sequential" processing since they are still processed in their own independent thread.

namirfa:
@fhcarter 
Updated the README to clarify the "parallel vs sequential" issue.

PR #188, opened 2019-09-13
OPC UA: Upgrade connector to a more modern Milo (SDK) version

fhcarter:
Upgraded Milo version used to version 0.3.3.  As of version 0.3 (.0?), the API changed a bit (it is still version 0...), so some code changes were necessary. Whilst there, updated the set of public OPC UA servers against which we test connections (and reported some bugs to them where appropriate).

Generalized our _loopback address reported by discovery handling_ to fix up any discovered addresses that include loopback or unresolvable or unreachable addresses.  This is, apparently, as per spec.  The spec says that the server need not be aware of how it can be reached, so the client must fix up URLs by using the host used for discovery.  So we now handle that..


fhcarter:
Note -- may hold off on merging this pending Olivier's trying at at the customer.  His call -- I can go either way.

PR #194, opened 2020-01-08
Improve returned results for object rec.

### LINKED ISSUE #193: ObjectRec: Roughly half of recognitions MIA
In evaluation of the YOLO results, the open source code we use puts results into a priority queue ordered by confidence -- only putting things in which are above the required confidence level. It then goes through and eliminates duplicate discoveries (of which there are usually a lot -- nature of the beast). The loop it uses is based (in part) on the queue's size compared to loop iterations, but the loop uses a priorityQueue.poll() call. This call removes an entry from the queue, reducing the queue's size, so only about half of the results are processed.

Fix is a simple change to use only the queue's size compared to zero.

fhcarter:
Fixes #193 

Update YOLOClassifier to process all the recognitions.  It was previously limited in some cases, and tended to process only half of them then.

Update tests to take this into account.

Ancillary: Add threshold to null suppression tests.  Either _because it does_ or because of the aforementioned change, these tests were failing more often (it tended to recognize a bed in the corner of the picture with 50% confidence).  Set threshold to 60% to mute these spurious failures.

### DIFF
--- a/objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
+++ b/objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
@@ -1797,16 +1797,36 @@ public class TestYoloProcessor extends NeuralNetTestBase {
 
     // ================================================= Helper functions =================================================
 
-    String imageResultsAsString = "[{\"confidence\":0.8445639, \"location\":{\"top\":255.70024, \"left\":121.859344, \"bottom\":372.2343, "
-            + "\"right\":350.1204}, \"label\":\"keyboard\"}, {\"confidence\":0.7974271, \"location\":{\"top\":91.255974, \"left\":164.41359, "
-            + "\"bottom\":275.69666, \"right\":350.50714}, \"label\":\"tvmonitor\"}]";
+    String imageResultsAsString = "[" +
+            "{\"confidence\":0.8445639, " +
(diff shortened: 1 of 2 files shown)


objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -1748,6 +1748,7 @@ public void testSuppressNullValuesHelper(boolean suppressNullValues) throws Inte
             for (int i = 0; i < responseBody.size(); i++) {
                 JsonObject resultObject = (JsonObject) responseBody.get(i);
                 assert resultObject.get("results") instanceof JsonArray;
+                System.out.println("Results: " + resultObject.get("results"));
```
namirfa: Might want to remove this `System.out.println()`

objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -1878,7 +1899,9 @@ void verifyProcessing(YoloProcessor ypImageSaver, String resultsAsString) throws
     }
 
     void resultsEquals(List<Map<String, ?>> list, List<Map> expectedRes) {
-        assert list.size() == expectedRes.size();
+        System.out.println("list: " + list);
```
namirfa: Ditto.

PR #211, opened 2020-11-06
Update build.gradle

olivier-bernard:
Modified the inclusion of files to enable HIKVISION_LOC to be the parent directory of HikVision SDK and include the relevent files in that directory.

Working on Linux, but need to check if this modification work on Linux


fhcarter (commented):
OK by me but you should really check with @sagizagagi.  HikVision is still his branch

PR #217, opened 2020-12-10
Add SDK support for getting base configuration information

### LINKED ISSUE #214: EXTJSDK: Build standard mechanism for getting config data into SDK
Add standard mechanism for reading the (probably) source.config file and making results available.  This should be added to the SDK, and it's use (amongst Vantiq-developed connectors) standardized.  Helpful in the deployment project on the server side.

fhcarter:
Fixes #214 

Add Util class to ExtJSdk to get config properties from $CWD/serverConfig/server.config.  Convert most, if not all, sources to use it.

Converted UPD source to pick up properties instead of JSON file.  Since it already converted the JSON file to a Map, I just took advantage of the fact that a Java Properties file IS a Map and used that.  Seems to work OK, at least in tests.

I may need to do more testing on these things. None of the connectors actually have tests that verify "real" configurations and none of them set up their source implementation types.  So it's a highly manual process right now.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Wed Dec 09 14:41:33 PST 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
-zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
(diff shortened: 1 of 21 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Dec 09 14:41:33 PST 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
```
namirfa, line 3: Just reorganizing things here?

namirfa (approved):
LGTM

gradle/wrapper/gradle-wrapper.properties
fhcarter: I think Intellij did something.  I can remove it since it's not really a change. But maybe I'll leave it so IntelliJ doesn't constantly fight with us.

PR #218, opened 2020-12-15
Using gradle to build docker images

namirfa:
Uses the `dcompose` plugin to build docker images for the connectors.

The developer will need to configure a `Dockerfile` for their connector, and add the proper `dcompose` task to their connector's `build.gradle` file.

This is not fully completed, it just shows an example of this working for the JDBC Connector. We will need to add similar code for each of the other connectors, which I can do next.

There are two things that still need to be completed:
1. Checking for the existence of Docker Hub credentials in the gradle.properties file, along with a Docker Hub registry to which we should push the image. Currently the `repository` portion of the `dcompose` task is just creating the image with that name in my local Docker Hub registry.
2. Forcing the `dcompose` task to depend on the default `assemble` task. This was not working when I tried it, though of course there must be a way to get this done.

After building the image with this tool, I manually pushed it to my public Docker Hub repository. I then ran the Connector test in `K8sClusterTests` using the debugger and was able to pulish/query the connector running in my local K8s cluster. Of course there was no DB running in the container so all the requests failed, but the connector received them all and attempted to send them along to the nonexistent DB.

### DIFF
--- a/CSVSource/build.gradle
+++ b/CSVSource/build.gradle
@@ -78,31 +78,3 @@ tasks.withType(Test) {
         systemProperty 'EntFullFilePath', rootProject.findProperty('EntFullFilePath')?:"b.csv"
     }
 }
-
-// Using gradle-dcompose-plugin to build CSVSource docker image
-dcompose {
-    if (rootProject.hasProperty('dockerRegistry')) {
-        def registryName = rootProject.getProperty('dockerRegistry')
-        csvSource {
-            buildFiles = project.copySpec {
-                from "$projectDir/src/main/docker/"
-                from "$projectDir/build/distributions/CSVSource.tar"
-            }
-            if (rootProject.hasProperty('csvImageTag')) {
-                repository = registryName + '/csv-source' + rootProject.getProperty('csvImageTag')
-            } else {
-                repository = registryName + '/csv-source'
-            }
-            logBuildStatus = true
-            buildLogFile = file("$buildDir/csvDockerBuildLog.file")
-        }
-
-        // The credentials here should be stored in gradle.properties.
-        dockerClientConfig = {
-            withRegistryUsername dockerRegistryUser
-            withRegistryPassword dockerRegistryPassword
-        }
-    }
-}
-
-buildCsvSourceImage.dependsOn assemble
--- a/CSVSource/build.gradle
+++ b/CSVSource/build.gradle
@@ -78,3 +78,31 @@ tasks.withType(Test) {
         systemProperty 'EntFullFilePath', rootProject.findProperty('EntFullFilePath')?:"b.csv"
     }
 }
+
+// Using gradle-dcompose-plugin to build CSVSource docker image
+dcompose {
+    if (rootProject.hasProperty('dockerRegistry')) {
+        def registryName = rootProject.getProperty('dockerRegistry')
+        csvSource {
+            buildFiles = project.copySpec {
+                from "$projectDir/src/main/docker/"
+                from "$projectDir/build/distributions/CSVSource.tar"
+            }
+            if (rootProject.hasProperty('csvImageTag')) {
+                repository = registryName + '/csv-source' + rootProject.getProperty('csvImageTag')
+            } else {
+                repository = registryName + '/csv-source'
+            }
+            logBuildStatus = true
+            buildLogFile = file("$buildDir/csvDockerBuildLog.file")
+        }
+
+        // The credentials here should be stored in gradle.properties.
+        dockerClientConfig = {
+            withRegistryUsername dockerRegistryUser
+            withRegistryPassword dockerRegistryPassword
+        }
+    }
+}
+
+buildCsvSourceImage.dependsOn assemble
--- a/README.md
+++ b/README.md
@@ -359,16 +359,17 @@ The connectors in this repository contain `gradle` tasks that can be used to bui
 push those images to the developer's Docker Registry. In order to use these tasks, the developer must include the 
 following configuration option in their `gradle.properties` file, along with some optional parameters:
 
-*   `dockerRegistry`: Required. The name of the registry to which the image should be pushed.
+*   `dockerRegistry`: Required. The name of the registry to which the image should be pushed (i.e. `docker.io`,
+`quay.io`, etc.).
+*   `pathToRepo`: Required. The path to the docker repository. This is typically the `namespace` portion of registry
+URIs that follow the `registry/namespace/repo:tag` structure, but each registry can very. This value should begin and
+end with "/" (i.e. `pathToRepo=/vantiq/`).
 *   `dockerRegistryUsername`: Optional. The username used for authenticating with the given docker registry.
 *   `dockerRegistryPassword`: Optional. The password used for authenticating with the given docker registry.
 *   `imageTag`: Optional. The tag used when pushing the image. If not specified, the tag will default to "latest".
 *   `repositoryName`: Optional. The name of the repository in the registry to which the image should be pushed. If not
 specified, the default repository will be the connector's name (i.e. "jdbc-source", "jms-source", 
 "objectrecognition-source", etc.).
-*   `useDockerHub`: Optional. A boolean flag that must be set to true if the docker registry is Docker Hub. This is 
-necessary in order to properly authenticate with Docker Hub. If the repository in question is public, then this can be 
-ignored.
 *   `connectorSpecificInclusions`: Optional. The path to a directory of files that need to be included in the image. 
 These can then be referenced and used by the Dockerfile.
 
--- a/README.md
+++ b/README.md
@@ -362,8 +362,7 @@ following configuration option in their `gradle.properties` file, along with som
 *   `dockerRegistry`: Required. The name of the registry to which the image should be pushed (i.e. `docker.io`,
 `quay.io`, etc.).
 *   `pathToRepo`: Required. The path to the docker repository. This is typically the `namespace` portion of registry
-URIs that follow the `registry/namespace/repo:tag` structure, but each registry can very. This value should begin and
-end with "/" (i.e. `pathToRepo=/vantiq/`).
+URIs that follow the `registry/namespace/repo:tag` structure, but each registry can vary, (i.e. `pathToRepo=/vantiq/`).
 *   `dockerRegistryUsername`: Optional. The username used for authenticating with the given docker registry.
 *   `dockerRegistryPassword`: Optional. The password used for authenticating with the given docker registry.
 *   `imageTag`: Optional. The tag used when pushing the image. If not specified, the tag will default to "latest".
--- a/README.md
+++ b/README.md
@@ -369,6 +369,8 @@ specified, the default repository will be the connector's name (i.e. "jdbc-sourc
 *   `notDockerHub`: Optional. A boolean flag that must be set to true if the docker registry is not Docker Hub. This is 
 necessary in order to properly authenticate with the given registry. If the repositories are public, then this can be 
 ignored.
+*   `connectorSpecificInclusions`: Optional. The path to a directory of files that need to be included in the image. 
+This data can then be referenced and used by the Dockerfile.
 
 With the required properties in place, the tasks can then be executed as follows:
 
--- a/README.md
+++ b/README.md
@@ -366,11 +366,11 @@ following configuration option in their `gradle.properties` file, along with som
 *   `repositoryName`: Optional. The name of the repository in the registry to which the image should be pushed. If not
 specified, the default repository will be the connector's name (i.e. "jdbc-source", "jms-source", 
 "objectrecognition-source", etc.).
-*   `notDockerHub`: Optional. A boolean flag that must be set to true if the docker registry is not Docker Hub. This is 
-necessary in order to properly authenticate with the given registry. If the repositories are public, then this can be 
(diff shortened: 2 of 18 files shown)


namirfa:
Of the two remaining items, the first one could theoretically be considered "optional". That is, we don't necessarily need to automate the image pushing. It would be a nice feature to have but could be part of "v2" or added in a second pass. 

The more critical piece is making sure that `assemble` is run first, since we need to pull the `.tar` file's contents from the build directory.

jdbcSource/build.gradle
```diff
@@ -84,4 +84,18 @@ dependencies {
     // Used to create a connection pool if asynchronous processing has been specified for publish/query handlers
     compile group: 'com.zaxxer', name: 'HikariCP', version: '3.3.1'
 
+}
+
+// Using gradle-dcompose-plugin to build jdbcSource docker image
+dcompose {
```
fhcarter, line 90: You can stick in something like:

```
ext {
    if (rootProject.hasProperty('dockerRepo')) {
        dockerRepo = rootProject.getProperty('dockerRepo')
    } else {
        dockerRepo = 'missing/missing'
    }
}
```

to get the docker repo.  Then, change `repository - ...` below to `repository - "$dockerRepo"`

fhcarter: Better will be to skip the task if the repo isn't set.  But I'll have to think about that more...

jdbcSource/build.gradle
```diff
@@ -84,4 +84,18 @@ dependencies {
...
+    jdbcSource {
+        buildFiles = project.copySpec {
+            from "$projectDir/src/main/docker/"
+            from "$projectDir/build/distributions/jdbcSource.tar"
+        }
+        repository = 'namirfa/namir-testing:jdbc'
```
fhcarter, line 96: I think I'd prefer not to have this default to pushing to your repo.  If you merge, it goes into master so anyone playing with it...

jdbcSource/build.gradle
```diff
@@ -84,4 +84,18 @@ dependencies {
...
+        logBuildStatus = true
+        buildLogFile = file("$buildDir/jdbcDockerBuildLog.file")
+
```
fhcarter, line 99: empty line.

fhcarter (changes requested):
Very nice.

I'd prefer that we didn't push to your repo since that'll likely break if we do this for real (that is, if you push to master).  I have hints in the comments to make all that work.

Also have a somewhat tested fix for the `dependsOn` stuff.

namirfa:
There's a lot of recycled code but I don't know if there's a way to generalize it- I'm not familiar enough with gradle-isms to know how to pull that stuff out into some sort of utility. Let me know what you think.

CSVSource/build.gradle
```diff
@@ -78,3 +78,31 @@ tasks.withType(Test) {
...
 }
+
+// Using gradle-dcompose-plugin to build CSVSource docker image
+dcompose {
+    if (rootProject.hasProperty('dockerRegistry')) {
+        def registryName = rootProject.getProperty('dockerRegistry')
+        csvSource {
+            buildFiles = project.copySpec {
+                from "$projectDir/src/main/docker/"
+                from "$projectDir/build/distributions/CSVSource.tar"
```
fhcarter, line 89:
```suggestion
                from "$projectDir/build/distributions/${project.name}.tar"
```

CSVSource/build.gradle
```diff
@@ -78,3 +78,31 @@ tasks.withType(Test) {
...
+            }
+            if (rootProject.hasProperty('csvImageTag')) {
+                repository = registryName + '/csv-source' + rootProject.getProperty('csvImageTag')
```
fhcarter, line 92: Something similar.  If you need the name in this format, you can mostly do something like

```suggestion
                repository = registryName + "${project.name - 'Source' + '-source}.toLowercase()" + rootProject.getProperty('imageTag')
```

CSVSource/build.gradle
```diff
@@ -78,3 +78,31 @@ tasks.withType(Test) {
...
+// Using gradle-dcompose-plugin to build CSVSource docker image
+dcompose {
+    if (rootProject.hasProperty('dockerRegistry')) {
+        def registryName = rootProject.getProperty('dockerRegistry')
+        csvSource {
+            buildFiles = project.copySpec {
+                from "$projectDir/src/main/docker/"
+                from "$projectDir/build/distributions/CSVSource.tar"
+            }
+            if (rootProject.hasProperty('csvImageTag')) {
```
fhcarter, line 91: Make the image tag generic.  If someone wants to publish everything with different tags, they can do 20 publishes.

```suggestion
            if (rootProject.hasProperty('imageTag')) {
```

CSVSource/build.gradle
```diff
@@ -78,3 +78,31 @@ tasks.withType(Test) {
...
+                repository = registryName + '/csv-source' + rootProject.getProperty('csvImageTag')
+            } else {
+                repository = registryName + '/csv-source'
+            }
```
fhcarter, line 95: Probably better to take the above construction & save it in a variable rather than redo it.

CSVSource/build.gradle
```diff
@@ -78,3 +78,31 @@ tasks.withType(Test) {
...
+            logBuildStatus = true
+            buildLogFile = file("$buildDir/csvDockerBuildLog.file")
+        }
```
fhcarter, line 98: Similar -- though here since they all go into their own build areas, can't it just be `dockerBuildLog.txt` (`.txt` rather than `.file` unless there's a good reason for that.   Then double clicking on it does something reasonable.

objectRecognitionSource/build.gradle
```diff
@@ -305,3 +305,38 @@ task copyTestResources(type: Copy, dependsOn: [configurations.testRuntime, confi
...
+        objectRecognitionSource {
+            buildFiles = project.copySpec {
+                from "$projectDir/src/main/docker/"
+                from "$projectDir/build/distributions/objectRecognitionSource.tar"
+                if (rootProject.hasProperty('objectRecognitionModelFiles')) {
+                    def modelLocation = rootProject.getProperty('objectRecognitionModelFiles')
+                    from "$modelLocation"
+                } else {
+                    from "$projectDir/build/models"
+                }
```
fhcarter, line 322: Here, you can either keep his one separate OR you can define a closure that does the model stuff. (`sourceSpecificAdditiions`} and if it's present, invoke it.  Then other things can do stuff similarly if necessary.

The generic one is invoked if present locally or something (that is `if project.hasProperty` kind-a thing

jmsSource/build.gradle
```diff
@@ -71,3 +71,31 @@ dependencies {
...
+
+        // The credentials here should be stored in gradle.properties.
+        dockerClientConfig = {
+            withRegistryUsername dockerRegistryUser
+            withRegistryPassword dockerRegistryPassword
+        }
+    }
+}
+
+buildJmsSourceImage.dependsOn assemble
```
fhcarter, line 101: Have to check if these get properly named if included in an `allProjects` or `subProjects`, but I think they will.

fhcarter (approved):
LGTM, I think.

I think you can do the generalization by using expressions based on the project name to name/fetch the artifacts.  That is, for example, rather than using the text `CSVSource`, use `"${project.name}"`.  Similarly, if you want repos named csv-source, you have to do some _string arithmetic_ on the name (see below -- I think I got it right -- close, anyway).  And make the imageTag a single property -- makes things easier.

README.md
```diff
@@ -351,4 +351,63 @@ all possible options,
...
+following configuration option in their `gradle.properties` file, along with some optional parameters:
+
+*   `dockerRegistry`: Required. The name of the registry to which the image should be pushed.
+*   `dockerRegistryUsername`: Optional. The username used for authenticating with the given docker registry.
+*   `dockerRegistryPassword`: Optional. The password used for authenticating with the given docker registry.
+*   `imageTag`: Optional. The tag used when pushing the image. If not specified, the tag will default to "latest".
+*   `repositoryName`: Optional. The name of the repository in the registry to which the image should be pushed. If not
+specified, the default repository will be the connector's name (i.e. "jdbc-source", "jms-source", 
+"objectrecognition-source", etc.).
+*   `notDockerHub`: Optional. A boolean flag that must be set to true if the docker registry is not Docker Hub. This is
```
fhcarter, line 369: Can't we tell some other way?  (I don't like this name).  Can't we just look at $dockerRegistry?

if we have to use it, can we reverse it to be `useDockerHub`?

README.md
```diff
@@ -351,4 +351,63 @@ all possible options,
...
+With the required properties in place, the tasks can then be executed as follows:
+
+### Build Image
+
+From the root directory of this repo, run the following command (this example builds the JDBC Connector)
+```
+./gradlew jdbcSource:buildConnectorImage
+```
+
+You can also use the camel-cased version:
```
fhcarter, line 384: we don't need to say this part, I don't think.  That's gradle stuff so we don't want to require it.

build.gradle
```diff
@@ -71,6 +72,51 @@ subprojects {
...
+            if (rootProject.hasProperty('repositoryName')) {
+                repositoryName = rootProject.getProperty('repositoryName')
+            }
+
+            connector {
+                buildFiles = project.copySpec {
+                    from "$projectDir/src/main/docker/"
+                    from "$projectDir/build/distributions/${project.name}.tar"
+                    if (rootProject.hasProperty('connectorSpecificInclusions')) {
+                        def inclusionsDir = rootProject.getProperty('connectorSpecificInclusions')
```
fhcarter, line 92: Do we need to say where they go?  Should that be another property. Or is this handled in the docker file?

build.gradle
```diff
@@ -71,6 +72,51 @@ subprojects {
...
+            def repositoryName = "${project.name - 'Source' + '-source'}".toLowerCase()
+            if (rootProject.hasProperty('repositoryName')) {
+                repositoryName = rootProject.getProperty('repositoryName')
+            }
+
+            connector {
+                buildFiles = project.copySpec {
+                    from "$projectDir/src/main/docker/"
+                    from "$projectDir/build/distributions/${project.name}.tar"
+                    if (rootProject.hasProperty('connectorSpecificInclusions')) {
```
fhcarter, line 91: Here, it seems like this could be something that the local project has, too.  For example, some connector could want some standard document included somewhere for and may define that in it's ext props?

build.gradle
```diff
@@ -71,6 +72,51 @@ subprojects {
     apply plugin: 'idea'
     apply plugin: 'eclipse'
     apply plugin: 'java'
+    apply plugin: 'com.chrisgahlert.gradle-dcompose-plugin'
+
+    // Using gradle-dcompose-plugin to build/push docker images
+    dcompose {
+        if (rootProject.hasProperty('dockerRegistry')) {
```
fhcarter, line 79: Here, if I'm reading right, we do nothing if the `dockerRegistry` isn't defined.  Don't we want to just not publish here?  That is, we'd have everything build except the client closures?

namirfa:
@fhcarter Fixed per our discussion.

build.gradle
```diff
@@ -71,6 +72,52 @@ subprojects {
...
+            }
+
+            // The credentials here should be stored in gradle.properties.
+            if (isDockerHub) {
+                dockerClientConfig = {
+                    withRegistryUsername dockerRegistryUser
+                    withRegistryPassword dockerRegistryPassword
+                }
+            } else {
+                registry(registryName) {
```
fhcarter, line 113: Here, registry name is something like `quay.io/fred` or `/fred`, where the last one means to default to Docker Hub.  Correct?

objectRecognitionSource/build.gradle
```diff
@@ -304,4 +304,4 @@ task copyTestResources(type: Copy, dependsOn: [configurations.testRuntime, confi
     into "$buildDir/testResources/"
 }
 
-test.dependsOn copyModels, copyTestResources
+test.dependsOn copyModels, copyTestResources
```
fhcarter, line 307: do we need a `buildConnectorImage.dependsOn copyModels`?

fhcarter (approved):
A question or two but looks OK, I think.

I actually think I have a way to avoid `useDockerHub`, but I'll play with it & update at a later time.

objectRecognitionSource/build.gradle
namirfa: I removed the code that copies the testing models into the OR connector by default, so we shouldn't need it. If we want to add that feature back then we'd need it.

fhcarter: OK -- missed that.

README.md
```diff
@@ -351,4 +351,54 @@ all possible options,
...
+
+## Building Docker Images
+
+The connectors in this repository contain `gradle` tasks that can be used to build Docker Images for each connector and 
+push those images to the developer's Docker Registry. In order to use these tasks, the developer must include the 
+following configuration option in their `gradle.properties` file, along with some optional parameters:
+
+*   `dockerRegistry`: Required. The name of the registry to which the image should be pushed (i.e. `docker.io`,
+`quay.io`, etc.).
+*   `pathToRepo`: Required. The path to the docker repository. This is typically the `namespace` portion of registry
```
fhcarter, line 364: It's fine & we can fix later.  I would remove the _must begin & end with '/'_ stuff and add as necessary.  But we can fix this later. This one's churned enough.

fhcarter (approved):

PR #220, opened 2021-01-05
Add ability to pickup credentials from environment variable

### LINKED ISSUE #219: EXTJSDK: Allow ability to pick up credentials, etc.
From other places (files, environment variables), etc.

namirfa:
Fixes #219 

Enables the ExtSrc SDK to pickup the `authToken` property if specified in an environment variable named `CONNECTOR_AUTH_TOKEN`. If an authToken is included in the `server.config` file, that value will be given precedence over the environment variable.

### DIFF
--- a/extjsdk/src/test/java/io/vantiq/extjsdk/TestUtils.java
+++ b/extjsdk/src/test/java/io/vantiq/extjsdk/TestUtils.java
@@ -11,6 +11,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Properties;
 
+import static org.junit.Assume.assumeTrue;
+
 
 public class TestUtils  {
 
@@ -21,6 +23,15 @@ public class TestUtils  {
     public static final String AUTH_TOKEN_PROP = "authToken";
     public static final String OTHER_PROP = "otherProperty";
 
+    public static String FAKE_TOKEN2 = "xxxx====xxxx";
+    public static String SECRET_CREDENTIALS = "CONNECTOR_AUTH_TOKEN";
+    String envVarAuthToken;
+
+    @Before
+    public void setup() {
+        envVarAuthToken = System.getenv(SECRET_CREDENTIALS);
+    }
+
     @Test
     public void testGetConfigAlone() throws Exception {
         BufferedWriter bw = null;
@@ -30,9 +41,9 @@ public class TestUtils  {
             f = new File(p.toString());
             f.deleteOnExit();
             
-            bw = fillProps(p);
+            bw = fillProps(p, true);
            
-            checkProps();
+            checkProps(false);
         } finally {
             if (bw != null) {
                 bw.close();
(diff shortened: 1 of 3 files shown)


extjsdk/src/test/java/io/vantiq/extjsdk/TestUtils.java
```diff
@@ -57,9 +68,53 @@ public void testGetConfigInDir() throws Exception {
...
+                envVarAuthToken != null && envVarAuthToken.equals(FAKE_TOKEN));
+        doEnvVarTests(false);
+    }
+
+    @Test
+    public void testGetEnvVarOverwrite() throws Exception {
+        assumeTrue("\"CONNECTOR_AUTH_TOKEN\" environment variable must be set equal to \"xxxx====xxxx\"",
+                envVarAuthToken != null && envVarAuthToken.equals(FAKE_TOKEN2));
+        doEnvVarTests(true);
+    }
```
namirfa, line 101: Since the environment variable can only be set to one value, one of these tests will always be ignored. Both pass if their respective correct value is specified.

extjsdk/src/test/java/io/vantiq/extjsdk/TestUtils.java
```diff
@@ -57,9 +68,53 @@ public void testGetConfigInDir() throws Exception {
...
+
+    @Test
+    public void testGetEnvVar() throws Exception {
+        assumeTrue("\"CONNECTOR_AUTH_TOKEN\" environment variable must be set equal to \"xxxx====\"",
+                envVarAuthToken != null && envVarAuthToken.equals(FAKE_TOKEN));
+        doEnvVarTests(false);
+    }
+
+    @Test
+    public void testGetEnvVarOverwrite() throws Exception {
```
fhcarter, line 97: How do these tests get run? It appears to be assumed that the user will manually set the various values? It seems like we should create some grade test tasks that set these values as expected so things get tested regardless of the caller's experience.

There are ways in gradle to set environment variables.

namirfa: Sure that makes sense, I'll add it.

namirfa:
@fhcarter 

The `test` task in the `extjsdk/build.gradle` file now sets the `CONNECTOR_AUTH_TOKEN` var to `xxxx====` if it isn't already specified.

I changed the logic around to prioritize the authToken value set in the `server.config` file since I think that makes more sense, and also makes the testing more straightforward because we don't need to unset the environment variable after testing so that the other TestUtils tests will pass in the next run. Overall I think it's just a bit more straightforward.

fhcarter (approved):
LGTM -- thanks

rblumer4:
@namirfa, can you update the PR description with the final behavior?   thanks!

namirfa:
> @namirfa, can you update the PR description with the final behavior? thanks!

No problem, I can do that now! Was waiting to merge this while doing some corresponding server work, but I might just update the PR and merge it now instead.

PR #226, opened 2021-01-16
Fixing reconnectHandler issue in connectors, and adding ability in SDK to queue up messages that failed to send while reconnecting, and resend later

### LINKED ISSUE #222: JDBC Connector: Timeout Exception when reconnectHandler is triggered
When the reconnectHandler is triggered for the JDBC Connector, it fails to initially reconnect to the Vantiq server in the first 10 seconds which leads to the following error:
```
ERROR io.vantiq.extsrc.jdbcSource.JDBCCore#JDBC1 - Could not reconnect to source within 10 seconds: 
java.util.concurrent.TimeoutException: null
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1784) ~[?:1.8.0_271]
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1928) ~[?:1.8.0_271]
```
Immediately after throwing the error the connector then successfully reconnects. The easiest way to reproduce this issue is to modify the source configuration associated with the connector.

There shouldn't need to be the initial failure before a successful reconnection.

namirfa:
Closes #222 

Fixes a bug in many of the connector's reconnectHandlers that threw an unnecessary TimeoutException. Also adds an enhancement that allows the SDK to queue up messages that would have failed to send while the connection between connector<--->Vantiq is down. The SDK now queues up these messages (up to the 25 most recent), and then sends them after a successful reconnection.

### DIFF
--- a/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
+++ b/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
@@ -53,29 +53,17 @@ public class CSVCore {
                 pollTimer.cancel();
                 pollTimer = null;
             }
-            
+
+            // Do connector-specific stuff here
             oConfigHandler.configComplete = false;
 
-            // Spin off a thread to handle doing the reconnect and error/log handling, this way the handler doesn't
-            // block the onMessage() handler in the ExtensionWebSocketListener from processing future messages
-            new Thread( () -> {
-                CompletableFuture<Boolean> success = client.connectToSource();
-                try {
-                    if ( !success.get(10, TimeUnit.SECONDS) ) {
-                        if (!client.isOpen()) {
-                            log.error("Failed to connect to server url '" + targetVantiqServer + "'.");
-                        } else if (!client.isAuthed()) {
-                            log.error("Failed to authenticate within 10 seconds using the given authentication data.");
-                        } else {
-                            log.error("Failed to connect within 10 seconds");
-                        }
-                        close();
-                    }
-                } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                    log.error("Could not reconnect to source within 10 seconds: ", e);
+            // Boiler-plate reconnect method, if reconnect fails then we call close()
+            CompletableFuture<Boolean> reconnectResult = client.doCoreReconnect();
+            reconnectResult.thenAccept(success -> {
+                if (!success) {
                     close();
                 }
-            }).start();
+            });
         }
     };
     
--- a/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
+++ b/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
@@ -57,7 +57,9 @@ public class CSVCore {
             // Do connector-specific stuff here
             oConfigHandler.configComplete = false;
 
-            // Boiler-plate reconnect method, if reconnect fails then we call close()
+            // Boiler-plate reconnect method- if reconnect fails then we call close(). The code in this reconnect
+            // handler must finish executing before we can process another message from Vantiq, meaning the
+            // reconnectResult will not complete until after we have exited the handler.
             CompletableFuture<Boolean> reconnectResult = client.doCoreReconnect();
             reconnectResult.thenAccept(success -> {
                 if (!success) {
--- a/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
+++ b/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
@@ -55,24 +55,27 @@ public class CSVCore {
             }
             
             oConfigHandler.configComplete = false;
-                        
-            CompletableFuture<Boolean> success = client.connectToSource();
-            
-            try {
-                if ( !success.get(10, TimeUnit.SECONDS) ) {
-                    if (!client.isOpen()) {
-                        log.error("Failed to connect to server url '{}'.", targetVantiqServer);
-                    } else if (!client.isAuthed()) {
-                        log.error("Failed to authenticate within 10 seconds using the given authentication data.");
-                    } else {
-                        log.error("Failed to connect within 10 seconds");
+
+            // Spin off a thread to handle doing the reconnect and error/log handling, this way the handler doesn't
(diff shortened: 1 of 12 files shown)


CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
```diff
@@ -55,24 +55,27 @@ public void handleMessage(ExtensionServiceMessage message) {
...
-                        log.error("Failed to connect to server url '{}'.", targetVantiqServer);
-                    } else if (!client.isAuthed()) {
-                        log.error("Failed to authenticate within 10 seconds using the given authentication data.");
-                    } else {
-                        log.error("Failed to connect within 10 seconds");
+
+            // Spin off a thread to handle doing the reconnect and error/log handling, this way the handler doesn't
+            // block the onMessage() handler in the ExtensionWebSocketListener from processing future messages
+            new Thread( () -> {
+                CompletableFuture<Boolean> success = client.connectToSource();
```
fhcarter, line 62: Could we put these forking the thread & work into a utility procedure or something in the SDK?  Then, each connector can handle reconnects by via

```
    // Do my connector special stuff
   client.handleReconnect() 

   // It may have to expose the configComplete or define an "andThen" type thing to set the configComplete when done.
```

Just seems like a lot of duplicated code...

extjsdk/build.gradle
```diff
@@ -81,6 +81,9 @@ dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
     // Used for uploading documents (images) to VANTIQ with VANTIQ SDK
     testCompile 'io.vantiq:vantiq-sdk:1.0.24'
+
+    // Used to create EvictingQueue for failed messages queue
+    implementation 'com.google.guava:guava:23.0'
```
fhcarter, line 86: Tweak, but if we're moving here from `compile` to `implementation`, should we change the ones above as well?

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -238,14 +260,15 @@ public void sendNotification(Object data) {
         if (data != null && (data.getClass().isArray() || data instanceof List)) {
             throw new IllegalArgumentException("Notifications cannot be lists or arrays.");
         }
+//        if (isConnected()) {
```
fhcarter, line 263: leftover...

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -256,7 +279,10 @@ public void sendNotification(Object data) {
                 outstandingNotifications.release();
                 throw e;
             }
+        } else {
+            failedMessageQueue.add(msg);
         }
+//        }
```
fhcarter, line 285: leftover

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -280,7 +306,11 @@ public void sendQueryResponseEmpty(String replyAddress) {
         Response response = new Response()
                 .status(QUERY_NODATA_CODE)
                 .addHeader(ExtensionServiceMessage.RESPONSE_ADDRESS_HEADER, replyAddress);
-        send(response);
+        if (isConnected()) {
```
fhcarter, line 309: Assuming the client caller can see `isConnected()`, is that sufficient to replace `configurationComplete`?

CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVCore.java
fhcarter, line 62: Are there any issues with degenerate conditions where we could run out of threads here?  Do we need to have only one of these open at a time? Or is that already handled?  Some `configInProgress` or something?

It's rare, so probably not necessary...

fhcarter (commented):
A comment or two -- it would be nice if we could combine al the thread forking into, at least, a single utility module somewhere.

namirfa:
@fhcarter 

Should be ready for a review now. I did not add the option to configure the `failedMessageQueue` size yet. Wondering if that's worth making configurable, or if we should wait till someone asks for it. It would be a straightforward-enough addition, but we'd need to tweak all the connector's configuration handlers to check for the optional property and use it accordingly.

extjsdk/README.md
```diff
@@ -154,6 +154,14 @@ Query errors are sent when a Query cannot be completed successfully. To send a Q
 *	The parameters are an Object array of that will be translated into strings and inserted into the message template
     based anywhere that `{<array index>}` is located.
 
+#### Sending Messages when Vantiq connection drops
+Over the lifetime of a connector, it is possible for the connection between itself and the Vantiq server to drop or get
```
fhcarter, line 158: might want to tweak the wording here so that it doesn't sound like a reconnect drops or restarts the connection.  I know what you mean, but it sounds like we're dropping the websocket connection here which we're not.

jmsSource/src/main/java/io/vantiq/extsrc/jmsSource/JMSCore.java
```diff
@@ -64,25 +64,16 @@
...
-                    if (!client.isOpen()) {
-                        log.error("Failed to connect to server url '" + targetVantiqServer + "'.");
-                    } else if (!client.isAuthed()) {
-                        log.error("Failed to authenticate within 10 seconds using the given authentication data.");
-                    } else {
-                        log.error("Failed to connect within 10 seconds");
-                    }
+            // Boiler-plate reconnect method, if reconnect fails then we call close()
+            CompletableFuture<Boolean> reconnectResult = client.doCoreReconnect();
+            reconnectResult.thenAccept(success -> {
```
fhcarter, line 72: It may be worth noting here (boilerplate comment) that `reconnectResult` can't complete until the handler completes.  When I read the code, I thought maybe `thenAccept` may be in a race condition with the async call above it.  I don't really think that's the case anyway, but, in fact, it can't happen anyway (due to the but we're fixing).

fhcarter (approved):
LG.  A couple of possible tweaks (not code).

Thx for diving in.

PR #232, opened 2021-01-26
Handling missing or extra slashes.

### LINKED ISSUE #231: Check for slashes when building/pushing images with gradle
We don't protect against missing "/" when creating the full `registry/namespace/repository:tag` definition to build and push the connector images.

We should check for their existence, and add them if not included.

namirfa:
Closes #231 

Add some protection to make sure that the full `registry/namespace/repository:tag` structure is built correctly. Makes sure that there are slashes in the appropriate spots.

Also tweaked the tag mechanism to make sure it actually works, the old way wasn't really working.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,17 +78,11 @@ subprojects {
     dcompose {
         if (rootProject.hasProperty('dockerRegistry') && rootProject.hasProperty('pathToRepo')) {
             def registryName = rootProject.getProperty('dockerRegistry')
-            registryName = registryName.endsWith('/') ? registryName - '/': registryName
-
             def pathToRepo = rootProject.getProperty('pathToRepo')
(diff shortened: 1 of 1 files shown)


namirfa:
@fhcarter this PR should reflect your issue comment now.

namirfa:
@fhcarter 

For some reason I can't get a regex to work and replace multiple slashes (anything >=2) with a single slash. The clunky alternative I have locally is to do something like this:

```
while (repositoryBase.contains('//')) {
    repositoryBase = repositoryBase.replaceAll('//', '/')
}
```

fhcarter (approved):
LGTM

fhcarter:
Just for completeness, I think it's OK to just deal with //'s. We're not trying to bullet proof things, just not be picky about whether path starts or ends with a slash. And just fixing it is easier than complaining.  And generally fixing it is easier than special casing startsWithSlash and endsWithSlash.

PR #235, opened 2021-03-07
Skip buildConnectorImage dependency when the task has not been created.

### LINKED ISSUE #234: Get rid of buildConnectorImage error message
Some folks have reported seeing an error similar to the following when trying to run gradle build or assemble tasks:
```
A problem occurred evaluating root project 'vantiq-extension-source'.

> Could not get unknown property 'buildConnectorImage' for project ':CSVSource' of type org.gradle.api.Project.
```

Currently, the workaround would be to comment out the following line in the top-level `build.gradle` file:
```
buildConnectorImage.dependsOn assemble
```
Users shouldn't have to remove this line in order to get things working properly.

fhcarter:
Fixes #234

This happens if the properties `dockerRegistry` and  `pathToRepo` are
missing.  This may often be the case when you are not intending to
or currently publishing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,7 +108,9 @@ subprojects {
             }
         }
     }
-    buildConnectorImage.dependsOn assemble
+    if (project.tasks.findByName('buildConnectorImage')) {
+        buildConnectorImage.dependsOn assemble
+    }
 
     // only run with code-coverage metrics if property is explicitly specified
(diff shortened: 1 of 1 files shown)


fhcarter:
Making this fix so that the repo is usable without edits.  Sorry to override the protocol but this is fatal without someone digging deeper.

PR #237, opened 2021-03-08
Provide default values for dockerUser & password

### LINKED ISSUE #236: Allow local docker build w/o username & password
We should provide a way to build local docker images w/o  the `dockerUserName` & `dockerUserPassword` properties.  The `dcompose` plugin seems deficient in that regard (to me).  

Seems like if those properties are not defined, we should just define them as empty.  The plugin will continue & can then do a local build.  Obviously,  you can't publish, but that should be OK.

fhcarter:
Otherwise, things tend to fail spectacularly even though you're just trying to create a local docker image.

Fixes #236

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,10 +62,23 @@ allprojects {
             cacheChangingModulesFor 0, 'seconds'
         }
     }
-
-
-
 }
+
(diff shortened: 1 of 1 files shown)


namirfa (approved):
LGTM

namirfa (approved):
Still looks good!

PR #244, opened 2021-03-15
Checks that semaphore is non-null before releasing

### LINKED ISSUE #242: Exception while closing socket
I faced an exception while closing the extension source connection when receiving reconnect message, that exception prevents receiving the message and the extension source started an infinite loop.

the exception caused because the outstandingNotification object is still null and the outstandingNotification.release is failed due to the null object. defending it from null helped a lot. 

file ExtensionWebSocketClient , line 301 
            outstandingNotifications.release();
add an if statment 
     if (outstandingNotifications != null)
            outstandingNotifications.release();

namirfa:
Fixes #242 

Adds checks to make sure the semaphore is non-null before trying to release. This was causing some issues for reconnect/close handlers, particularly when the `ExtensionWebSocketListener` called `client.acknowledgeNotification()`.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -275,15 +275,16 @@ public class ExtensionWebSocketClient {
         ExtensionServiceMessage msg = new ExtensionServiceMessage("");
         msg.fromMap(m);
         if (isConnected()) {
+            Semaphore localOutstandingNotifications = outstandingNotifications;
             try {
-                outstandingNotifications.acquire();
+                localOutstandingNotifications.acquire();
                 this.send(msg);
             } catch (InterruptedException ie) {
                 log.warn("Obtaining space to sent notifications was interrupted.", ie);
             } catch (Exception e) {
                 // If we get an exception during the send, we're unlikely to get a response so release now.
-                if (outstandingNotifications != null) {
-                    outstandingNotifications.release();
+                if (localOutstandingNotifications != null) {
+                    localOutstandingNotifications.release();
                 }
                 throw e;
             }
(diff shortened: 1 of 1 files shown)


namirfa:
@fhcarter 

One question here: should we reinstantiate the `outstandingNotifications` if it is null? My concern is that I'm not totally sure who is nullifying it in the first place, that should only happen when the `stop()` method is called and I don't think that's happening.

That makes it tricky to decide where/if it should be reinstantiated- maybe in the `connectToSource` method.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -282,7 +282,9 @@ public void sendNotification(Object data) {
                 log.warn("Obtaining space to sent notifications was interrupted.", ie);
             } catch (Exception e) {
                 // If we get an exception during the send, we're unlikely to get a response so release now.
-                outstandingNotifications.release();
+                if (outstandingNotifications != null) {
```
fhcarter, line 285: Is there a race condition here whereby `isConnected()` is true when we check, but false (& thus oustandingNotifications has gone away?

Is this a case where we should have a local version of `outstandingNotifications` that we get then then check that?  Do we need to check the `outstandingNotifications` reference above?

That is
```
if (isConnected()) {
    localoutstandingNotifications = outstandingNoficiations
....
    if (localOuut...) {
        localist....acquire()
   }
....
....  
       lockout.release()
```

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -298,7 +300,9 @@ public void sendNotification(Object data) {
      * receipt of a response message.
      */
     void acknowledgeNotification() {
-        outstandingNotifications.release();
+        if (outstandingNotifications != null) {
+            outstandingNotifications.release();
```
fhcarter, line 304: also here, I think, but I don't recall all of what can go on at a time.

fhcarter (approved):
I think there's still a slight window here.  I think you can snarf a local copy (probably in both cases).  Then, use that to perform the release.

I don't think you need to get a new one -- assuming that `stop()` can only be unstopped by reconnecting, that should be what's required. It looks like it is created on class load, and on `stop()`.  Only?

PR #248, opened 2021-03-23
Removing model-copying from the Object Recognition Connector's Dockerfile

### LINKED ISSUE #247: Error when building objectRecognitionSource docker image
There is an error when running the `buildDockerImage` task for the objectRecognitionSource which is coming from the Dockerfile. Particularly, it's trying to copy in .meta/.pb/.labels files into the container by default that don't exist. The idea here was to copy in the generic model files that come from the `copyModels` task in that connector, but it's probably better to leave that up to the user and only include them via the `connectorSpecificInclusions`.

namirfa:
Closes #247 

Removing Dockerfile commands originally intended to copy .meta/.pb/.labels files downloaded by the `copyModels` task. Since the `buildConnectorImage` task isn't currently configured to depend on the `copyModels` task, we get an error when building the objectRec image.

In any case, it makes more sense to not include any model files by default and encourage the developer to use the `connectorSpecificInclusions` gradle property instead.

### DIFF
--- a/objectRecognitionSource/src/main/docker/Dockerfile
+++ b/objectRecognitionSource/src/main/docker/Dockerfile
@@ -1,9 +1,5 @@
 FROM openjdk:11-jre-slim
 RUN mkdir /app
-RUN mkdir /app/models
 ADD objectRecognitionSource.tar /app
-ADD *.meta /app/models/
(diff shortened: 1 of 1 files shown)


fhcarter (approved):
LGTM

PR #249, opened 2021-03-23
Removing references to deprecated syntax

### LINKED ISSUE #241: Remove all references to "WHEN MESSAGE ARRIVES ON <sourceName>"
This VAIL syntax has been deprecated, but we still include many examples using it. Those examples live both in the READMEs for the connectors, as well as the sample projects included in the resources.

It should instead be changed to `WHEN EVENT OCCURS ON "/sources/<sourceName>"`

namirfa:
Closes #241 

Removed all references to the `WHEN MESSAGE ARRIVES FROM <sourceName>` syntax, and replaced it with the supported version: `WHEN EVENT OCCURS ON "/sources/<sourceName>"`.

Needed to update most connector READMEs, 3 tests that create rules, and most of the connector example projects.


objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/neuralNet/TestYoloProcessor.java
```diff
@@ -85,7 +85,7 @@
     static final int KEYBOARD_CROPPED_HEIGHT = 125;
 
     // Used to test suppressNullValues
-    static final String NO_RECOGNIZED_OBJECTS_CAMERA_ADDRESS = "http://183.77.203.213:80/-wvhttp-01-/GetOneShot?image_size=640x480&frame_count=1000000000";
+    static final String NO_RECOGNIZED_OBJECTS_CAMERA_ADDRESS = "http://90.41.66.162:9000/mjpg/video.mjpg";
```
namirfa, line 88: This test was failing because the old link no longer exists.

fhcarter (approved):
LGTM -- Thanks

PR #252, opened 2021-04-20
Adds handling for K8s Readiness/Liveness TCP Probe

### LINKED ISSUE #215: EXTJSDK: Build in support for aliveness & readiness checks
... for connectors hosted in K8s environments

namirfa:
Closes #215 

Adds enhancements to the SDK to support K8s Readiness/Liveness TCP Probes. The TCP socket listener will listen on port 8000 by default, but this can be overridden by specifying the `tcpProbePort` in the `server.config` file.

The TestConnector was also enhanced to be able to set/unset its "health state" via publish/query requests. Not sure if we want to add this for all of the connectors, but that can be done in a separate issue/PR.

Need to add documentation for the new SDK features. Manually tested with my local minikube, and readiness/liveness checks all worked as expected.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -16,7 +16,6 @@ package io.vantiq.extjsdk;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 // WebSocket imports
-import groovy.json.internal.IO;
 import okhttp3.*;
 import okhttp3.ws.WebSocket;
 import okhttp3.ws.WebSocketCall;
@@ -205,7 +204,7 @@ public class ExtensionWebSocketClient {
      * readiness/liveness probe.
      */
     @SuppressWarnings("InfiniteLoopStatement")
-    synchronized public void declareHealthy() {
+    public synchronized void declareHealthy() {
         // Only reinitialize things if we have to, otherwise we just leave things running
         if (probeFuture != null && livenessSocket != null && !livenessSocket.isClosed()) {
             return;
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -20,6 +20,8 @@ import okhttp3.*;
 import okhttp3.ws.WebSocket;
 import okhttp3.ws.WebSocketCall;
 
+import java.io.IOException;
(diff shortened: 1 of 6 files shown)


extjsdk/src/main/java/io/vantiq/extjsdk/Utils.java
```diff
@@ -49,4 +49,27 @@ public static Properties obtainServerConfig(String fileName) {
...
+        File configFile = new File(SERVER_CONFIG_DIR, SERVER_CONFIG_FILENAME);
+        Properties properties = new Properties();
+        if (!configFile.exists()) {
+            configFile = new File(SERVER_CONFIG_FILENAME);
+            if (!configFile.exists()) {
+                return null;
+            }
+        }
+        properties.load(new FileReader(configFile));
+        String portString = properties.getProperty("tcpProbePort");
```
namirfa, line 68: This should be a constant var, not a string. I'll change in a future commit.

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -41,73 +43,109 @@
     /**
      * The code for a query response where more data will be sent.
      */
-    public static final int QUERY_CHUNK_CODE    = 100;
+    public static final int QUERY_CHUNK_CODE = 100;
```
namirfa, line 46: A few formatting cleanups at the top of this file.

fhcarter (changes requested):
At least one of the sider things is worth looking at.

I think that I'd like to see some notion of healthy/unhealthy rather than exposing the low-level TCP details. It seems like exposing these low-level details will (and has) caused issues with the connector-writer having to track too much internal SDK details rather than a more declarative mechanism.  Maybe there's a core level health setting that employed the future & TCP socket rather than each caller having to understand that.

namirfa:
@fhcarter 

Added fixes per your comments, ready for re-review.

fhcarter (approved):
A few minor issues to address but LGTM.

Some constants should be declared as such, a concurrency issue, and a test question.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
```diff
@@ -173,11 +196,9 @@ public void testProcessInvalidRequest() {
 
     @Test
     public void testExitIfConnectionFails() {
-        core.start(3);
```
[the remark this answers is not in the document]
namirfa: The `setup()` method is called before each test run and does the `core.start()`, so I removed it here.

PR #259, opened 2021-05-18
Adds ability for Test Connector to read raw bytes from files

### LINKED ISSUE #258: Enhance Test Connector for K8s LCM tests
Add a request parameter that signifies to the Test Connector to read raw bytes from a file without converting to string. Then base64 encode the data and send back to Vantiq.

Used in K8s LCM tests to ensure that mounting binary data works as intended, and also to test new `treatAsText` param.

namirfa:
Closes #258 

This enhancement adds an additional boolean parameter to Test Connector requests that, when specified as true, enables the  Test Connector to read data from files as raw bytes, and return them to Vantiq as a base64 encoded string.

This feature will be used for testing the K8s LCM tools in the Vantiq server code.

### DIFF
--- a/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
+++ b/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
@@ -12,7 +12,11 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeTrue;
 
-import java.util.*;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+import java.util.Base64;
 
 import org.junit.After;
 import org.junit.Before;
@@ -38,9 +42,9 @@ public class TestTestConnectorCore {
 
     @BeforeClass
     public static void checkFilesAndEnvVars() {
-        environmentVariable = System.getProperty("TestConnectorEnvVarName", null);
-        filename = System.getProperty("TestConnectorFilename", null);
-        fileContents = System.getProperty("TestConnectorFileContents", null);
+        environmentVariable = System.getProperty("TestConnectorEnvVarName");
+        filename = System.getProperty("TestConnectorFilename");
+        fileContents = System.getProperty("TestConnectorFileContents");
     }
 
     @Before
--- a/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
+++ b/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
@@ -150,7 +150,6 @@ public class TestTestConnectorCore {
         assert !fileString.isEmpty();
         assert !fileString.equals(fileContents);
         String decodedString = new String(Base64.getDecoder().decode(fileString.getBytes()));
-        decodedString = decodedString.replace(System.getProperty("line.separator"), ""); // Removing new lines
         assert decodedString.equals(fileContents);
     }
 
--- a/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
+++ b/testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
@@ -12,10 +12,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeTrue;
 
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import org.junit.After;
 import org.junit.Before;
@@ -35,6 +32,7 @@ public class TestTestConnectorCore {
 
     private static String environmentVariable;
     private static String filename;
+    private static String fileContents;
 
     private static final String NONEXISTENT_ENV_VAR = "anEnvironmentVariableThatIsUnlikelyToExist";
 
@@ -42,6 +40,7 @@ public class TestTestConnectorCore {
     public static void checkFilesAndEnvVars() {
         environmentVariable = System.getProperty("TestConnectorEnvVarName", null);
         filename = System.getProperty("TestConnectorFilename", null);
+        fileContents = System.getProperty("TestConnectorFileContents", null);
     }
 
     @Before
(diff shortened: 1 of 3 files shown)


namirfa:
@fhcarter Is this feature worth documenting?

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
```diff
@@ -12,10 +12,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeTrue;
 
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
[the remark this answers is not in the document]
fhcarter: I don't think we want to change to .* -- this is probably an IDE configuration but ...* is generally best avoided.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
```diff
@@ -35,13 +32,15 @@
...
     private static String filename;
+    private static String fileContents;
 
     private static final String NONEXISTENT_ENV_VAR = "anEnvironmentVariableThatIsUnlikelyToExist";
 
     @BeforeClass
     public static void checkFilesAndEnvVars() {
         environmentVariable = System.getProperty("TestConnectorEnvVarName", null);
         filename = System.getProperty("TestConnectorFilename", null);
+        fileContents = System.getProperty("TestConnectorFileContents", null);
```
fhcarter, line 43: just FYI, I think you can leave out the 2nd parameter & it'll default to null.  Usually you need that only to provide a non-null default.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
```diff
@@ -127,6 +126,30 @@ public void testProcessRequest() {
...
+
+        // Check that the data is base64 encoded version of what we expect
+        assert result.get("files") instanceof Map;
+        Map resultFiles = (Map) result.get("files");
+        assert resultFiles.get(filename) instanceof String;
+        String fileString = (String) resultFiles.get(filename);
+        assert !fileString.isEmpty();
+        assert !fileString.equals(fileContents);
+        String decodedString = new String(Base64.getDecoder().decode(fileString.getBytes()));
+        decodedString = decodedString.replace(System.getProperty("line.separator"), ""); // Removing new lines
```
fhcarter, line 149: Why are there any newlines?  if we put binary data in, there should be no newlines (unless they were in the binary data) on output.  The base64 encoded data may have newlines or other whitespace, but the decoding should remove them.  So I don't know why there's a removal of newlines after decoding.

Maybe this is because of the stringificiation above?

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
```diff
@@ -127,6 +126,30 @@ public void testProcessRequest() {
...
+        Map result = core.processRequest(request, null);
+
+        // Check that the data is base64 encoded version of what we expect
+        assert result.get("files") instanceof Map;
+        Map resultFiles = (Map) result.get("files");
+        assert resultFiles.get(filename) instanceof String;
+        String fileString = (String) resultFiles.get(filename);
+        assert !fileString.isEmpty();
+        assert !fileString.equals(fileContents);
+        String decodedString = new String(Base64.getDecoder().decode(fileString.getBytes()));
```
fhcarter, line 148: Why is this String as opposed to bytes?

fhcarter (changes requested):
Mostly OK.  Note about `getProperty()` -- you don't need to change anything, but when the 2nd parameter is null (the default default), just leave it out & use the single parameter overload.

Also, your IDE/project seems to have a low threshold for converting multiple includes to .*.  Generally, we don't want to do that.

The only real substantive issue is a question about why you're removing newlines from a base64-decoded byte stream?  I don't think that's correct. If it's necessary, I think something else is wrong.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
namirfa: The `fileString` is the base64 encoded version of the bytes, and then I test that things actually work by showing that the base64 version is not equal to the contents of the file, but once decoded they match up.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
namirfa: Fixed

fhcarter:
> @fhcarter Is this feature worth documenting?

Which feature?  I think we'd want to document ways to set up binary files.  Or are you asking about the test connector and/or it's readme.  Sure, why not -- someone else may want it or want to know what it means.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
namirfa: I couldn't quite figure out where things were going wrong.

I've created a test file on my machine that I used for this test. It just contains a few words. Initially this test kept on failing because somewhere along the road a `\n` is added to the end of the base64 encoded string. I figured it must be something related to the Mac OS adding a new line to the end of files, so for simplicities sake I just chose to remove that new line.

namirfa:
> > @fhcarter Is this feature worth documenting?
> 
> Which feature? I think we'd want to document ways to set up binary files. Or are you asking about the test connector and/or it's readme. Sure, why not -- someone else may want it or want to know what it means.

I just meant documenting the "rawBytes" parameter, I'll add it in now.

testConnector/src/test/java/io/vantiq/extsrc/testConnector/TestTestConnectorCore.java
fhcarter: I see what it's doing but you're passing strings as binary, I guess.  Why not just deploy some "real" binary data & check against that. So the fileString coming back is a base64 encoded string -- got that.  But I don't understand why `decodedString` is a `String`?  Why not just get the bytes & compare.

And, even with all that, why must we string the newlines out of the decoded stuff.  Decode(encode(orig)) == orig should always be true.

namirfa:
@fhcarter 

Got rid of the new line removal stuff. It was an issue with my machine adding a carriage return to the file. I just created the file from the command line with an `echo -n myTest > myFile.txt` and the test passed.

fhcarter:
OK, that makes sense.  For text files (cat - > file), it will properly terminate the file.  The issue is that these aren't text files, so creating them as such (as opposed to "real" binary files) is the issue.  

For the "real" test (K8sLiveConnector...), is should probably create binary data  to publish then use that for comparison.

fhcarter (approved):
LGTM

namirfa:
> OK, that makes sense. For text files (cat - > file), it will properly terminate the file. The issue is that these aren't text files, so creating them as such (as opposed to "real" binary files) is the issue.
> 
> For the "real" test (K8sLiveConnector...), is should probably create binary data to publish then use that for comparison.

The K8sLiveConnector test will deploy actual binary data. I was planning to create a `byte[]` of non UTF-8 characters to stress that everything is working and that we aren't using UTF-8 strings anywhere.

PR #264, opened 2021-06-07
Updating OkHttp to the latest version

### LINKED ISSUE #195: Java SDK:  Set up to handle ping's as required/desired
We've now seen a few cases where connectors that are not terribly active get timed out by the server.  Internally, they should handle reconnect, but it might be good to avoid doing this.

OkHttp3 (which we use for our WS implementation) can create a client that will automatically perform ping processing (`OkHttpClient.Builder()...pingInterval(duration)`).  This might require that we upgrade the OkHttp3 version we use (we currently use 3.4, that's up to 3.14 (I think), and 4.x).

We may want to do this as an option -- "active" connectors don't need it and it'll just increase communication.  But if there's a type of connector that gets work to do (or work that requires communication) rarely, then it may be a good thing for which to provide.

### LINKED ISSUE #204: Add keepAlive ping to connector SDK
Depending on the cloud provider, an idle connection between the connector and VANTIQ may get closed on some time interval, forcing a reconnection. In some cases, we've seen a bug on the VANTIQ side claiming a "connection already exists" for the given source. This issue seems to occur fairly randomly, and I have not been able to find a consistently reproducible case.

Adding some keepAlive functionality in the underlying connector SDK could prevent that scenario from ever arising, because the connection would never go idle. The VANTIQ server has the ability to respond to "ping" messages, so we'd only need to update the SDK code.

### LINKED ISSUE #260: Update our version of OkHttp
Update the OkHttp library to at least verson 3.10.0 so that we can pickup the `pingInterval` method. We can then offer SDK overrides to utilize the `pingInterval` if desired.

namirfa:
Closes #260 
Closes #204 
Closes #195

Updates the OkHttp library to the latest stable release. Since version 3.5, OkHttp has folded in the separate `-ws` package into the general okhttp one. The majority of these changes convert our old WebSocket implementations to use the newer version.

This PR also adds the ability to configure the OkHttp client to send pings to the Vantiq Server, which can be specified by including a `sendPings` property in the `server.config` file.

### DIFF
--- a/extjsdk/README.md
+++ b/extjsdk/README.md
@@ -18,8 +18,7 @@ ExtensionWebSocketListener to handle various message types.
 4.	Copy and connect extjsdk.jar to your project.
 5.	Include the following dependencies through Gradle, Maven, or whatever build manager you are using. Links to the
 dependencies on mvnrepository.com are included.
-	*	[com.squareup.okhttp3:okhttp Version 3.4.1](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/3.4.1)
-	*	[com.squareup.okhttp3:okhttp-ws Version 3.4.1](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp-ws/3.4.1)
+	*	[com.squareup.okhttp3:okhttp Version 4.9.1](https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp/4.9.1)
 	*	[com.fasterxml.jackson.core:jackson-databind Version 2.9.3](https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind/2.9.3)
     *   [org.slf4j:slf4j-api Version 1.7.25](https://mvnrepository.com/artifact/org.slf4j/slf4j-api/1.7.25)
 
--- a/extjsdk/README.md
+++ b/extjsdk/README.md
@@ -35,6 +35,7 @@ qualified class name, appended by a '#' and the name of the source they are asso
 
 ## <a name="serverConfig" id="serverConfig"></a>Connector Startup Configuration
 Generally, connectors need a minimum of three configuration properties at startup:
+
 *   `targetServer`: The URL of the Vantiq server to which the connector should connect.
 *   `authToken`: The access token that the connector will use to authenticate to the desired namespace.
 *   `source(s)`: The name of the source in the namespace (or in some cases a list of comma-separated source names) 
@@ -47,6 +48,7 @@ The SDK includes a utility method to retrieve the startup configuration document
 directory if it is not found in the `serverConfig` subdirectory (i.e. `<connectorWorkingDirector>/server.config`).
 
 In addition to the minimum configuration properties, the `server.config` file can also include the following property:
+
 *   `sendPings`: A boolean property that, if set to `true`, enables the SDK to send ping messages to the Vantiq Server. 
 The ping messages are handled by the underlying OkHttp library.
 
--- a/extjsdk/README.md
+++ b/extjsdk/README.md
@@ -46,6 +46,10 @@ The SDK includes a utility method to retrieve the startup configuration document
 `<connectorWorkingDirector>/serverConfig/server.config`. The SDK will also look for the file in the working connector 
 directory if it is not found in the `serverConfig` subdirectory (i.e. `<connectorWorkingDirector>/server.config`).
 
+In addition to the minimum configuration properties, the `server.config` file can also include the following property:
+*   `sendPings`: A boolean property that, if set to `true`, enables the SDK to send ping messages to the Vantiq Server. 
+The ping messages are handled by the underlying OkHttp library.
+
 For users who may not want to write the `authToken` property to a file because of its sensitive nature, the 
 `Utils.obtainServerConfig()` method will also search for this value in an environment variable named 
 `CONNECTOR_AUTH_TOKEN`. If the `authToken` is specified in the `server.config` document, that value will take precedence.
@@ -59,6 +63,7 @@ variable instead):
 authToken=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 sources=MySourceName
 targetServer=https://dev.vantiq.com/
+sendPings=true
 ```
 
 ## Program Flow
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
@@ -16,6 +16,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import okhttp3.WebSocket;
 import okhttp3.WebSocketListener;
 import okio.ByteString;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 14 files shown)


namirfa:
[Jenkins Build](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/21/)

fhcarter:
Shouldn't we add the `sendPing` documentation here?

```
targetServer=https://dev.vantiq.com/
```
								
[extjsdk/README.md](https://github.com/vantiq/vantiq-extension-sources/blob/4a1a713168cb2a6b8226d4edd03024556a101808/extjsdk/README.md#L61) (Line 61)

namirfa:
> Shouldn't we add the `sendPing` documentation here?
> 
> ```
> targetServer=https://dev.vantiq.com/
> ```
> 
> [extjsdk/README.md](https://github.com/vantiq/vantiq-extension-sources/blob/4a1a713168cb2a6b8226d4edd03024556a101808/extjsdk/README.md#L61) (Line 61)

Yep I'm adding that as we speak!

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -511,9 +505,21 @@ public boolean receivedMessage() {
         }
 
         @Override
-        public void sendPing(Buffer payload) throws IOException {}
+        public boolean close(int code, String reason) {
+            client.getListener().onClosed(null, code,reason);
```
fhcarter, line 509: Seems like instead of `null` for the WebSocket, we can pass in `this` can't we?  Seems like that would be cleaner, since it is a `WebSocket`.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -511,9 +505,21 @@ public boolean receivedMessage() {
...
+            return true;
+        }
+
+        //================================ Necessary to implement WebSocket ================================
+
+        @Override
+        public void cancel() {}
+        @Override
+        public long queueSize() {return 0;}
```
fhcarter, line 518: Same stuff about the multiline things to keep Sider happy.

namirfa:
@fhcarter
Documented. If it's alright with you, I'm going to create a separate issue for the TODO I added in the `Utils` class.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
```diff
@@ -511,9 +505,21 @@ public boolean receivedMessage() {
...
+        @Override
+        public Request request() {return null;}
```
fhcarter, line 520: ditto the sider stuff & multiline

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketListener.java
```diff
@@ -212,25 +209,13 @@ public void onOpen(WebSocket webSocket, okhttp3.Response response) {
      * Translate the received message and pass it on to the related handler. Additionally, updates the client about
      * successful authentications and source connections.
      *
-     * @param body  The {@link ResponseBody} containing the message received.
+     * @param webSocket The websocket on which the message was received.
+     * @param bodyBytes  The {@link ByteString} containing the message received.
      */
     @Override
-    public void onMessage(ResponseBody body) {
+    public void onMessage(WebSocket webSocket, ByteString bodyBytes) {
```
fhcarter, line 216: In the IDE, all the `on*()` calls seem to have a complaint from the IDE that your `@Override`s aren't annotated the same way as their super-methods.  Most are missing the `@NonNull`.

Probably better to keep these clean though it may cause issues with the tests that try to pass in `null` `WebSocket` instances a lot.  Probably better to just pass in the appropriate `FalseWebSocket`.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketListener.java
```diff
@@ -115,7 +116,7 @@ public void testSourceConnection() {
     public void testSourceConnectionFailFirst() {
         authenticate(true);
         
-        listener.onMessage(TestListener.errorMessage());
+        listener.onMessage(null, TestListener.errorMessage());
```
fhcarter, line 119: Sending the actual web socket here would be much preferable...

extjsdk/src/main/java/io/vantiq/extjsdk/Utils.java
```diff
@@ -84,4 +85,30 @@ public static Integer obtainTCPProbePort() {
...
+            } else {
+                return false;
+            }
+        } catch (IOException e) {
+            log.error("An error occurred while trying to retrieve the sendPings value.", e);
+            return false;
+        }
+    }
+
+    // TODO NAMIR - Add helper that returns the properties object since that code is repeated
```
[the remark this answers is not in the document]
fhcarter: Can be separate issue, I suppose. But since we support naming your config file, that support should extend here, too.

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketClient.java
fhcarter, line 518: ditto the sider stuff & multiline things

extjsdk/src/test/java/io/vantiq/extjsdk/TestExtensionWebSocketListener.java
fhcarter, line 119: May have to extend the client or listener to return the `FalseWebSocket` they use to pass in the right thing here.

extjsdk/README.md
```diff
@@ -47,6 +46,10 @@ The SDK includes a utility method to retrieve the startup configuration document
 `<connectorWorkingDirector>/serverConfig/server.config`. The SDK will also look for the file in the working connector 
 directory if it is not found in the `serverConfig` subdirectory (i.e. `<connectorWorkingDirector>/server.config`).
 
+In addition to the minimum configuration properties, the `server.config` file can also include the following property:
+*   `sendPings`: A boolean property that, if set to `true`, enables the SDK to send ping messages to the Vantiq Server.
```
fhcarter, line 50: To keep all the MD processors happy, usually a good idea to surround bullet lines with blank lines.

fhcarter (changes requested):
Mostly good.  I think there are a few places where it makes sense to clean up the sider warnings (keeps with our internal coding style anyway).  Also, I tend to think it'd be better in the tests to pass the websocket thru (rather than null).  Doesn't do that much, but allows the various `on` handlers to be properly declared (which the IDE is whining about anyway).

namirfa:
@fhcarter Added fixes per your comments.

[Latest Jenkins Build](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/22/)

Manually ran tests with server, which are all passing.

namirfa:
I did not add the `Utils` fix yet. Those methods should definitely accept a filename parameter, but I agree with your comment that it may be better to refactor those Utils in general to only read the `server.config` once, (and offer a `clear()` method as well). I'd like to make those changes in a separate PR if that's alright.

fhcarter:
I think that's OK, but we should try to do it soon.  It leaves the new stuff in a sort-of broken state (must used default file name) technically.  I don't think any of that is used yet, but still...

fhcarter (approved):
LGTM functionally.

There's one issue we should fix (`@NonNull` parameters) and a bunch of code style (missing/extra blank lines) issues.  But you can fix those & merge w/o another review.

Good to get the sider complaints done.  I'm sort of on a push to try & deal with those.  I wasn't impressed before, but some things are useful. (others not, but I'm still working on how to most easily disable the things we don't care about so we don't have to annotate all the files).

PR #274, opened 2021-07-16
Fixing outstandingNotifications NPE

### LINKED ISSUE #273: ExtjSDK - outstandingNotifications set to null causing NPE
In the case that the `ExtensionWebSocketClient` is instantiated -> `initiateFullConnection` is called -> then `close` is called, the next call to `initiateFullConnection` will leave the client in a state where it can no longer send notifications.

That is because the client's close/stop methods set the `outstandingNotifications` Semaphore to `null`, but the `initiateFullConnection` method is not responsible for reinstantiated the Semaphore. To avoid this, we should no longer set the Semaphore to `null`- we'll just reuse it.

namirfa:
Closes #273 

Removes the call to nullify outstandingNotifications. This means we'll no longer run into the issue that closing/reinitiating the SDK client's websocket will result in NPEs that prevent the user from sending notifications back to Vantiq.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -799,7 +799,6 @@ public class ExtensionWebSocketClient {
         // Saving and nulling before closing so EWSListener can know when it is closed by the client 
         WebSocket socket = webSocket;
         webSocket = null;
-        outstandingNotifications = null;
         if (socket != null) {
             try {
                 socket.close(1000, "Closed by client");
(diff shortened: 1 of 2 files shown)


fhcarter (approved):
LGTM

namirfa:
[Jenkins Build](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/27/)

namirfa:
@fhcarter 

There's a bunch of [OPCUA related failures](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/27/console) from Jenkins, all of which are passing on my machine.

Looks like some `Encrypt Private Key` related issues? Let me know if those seem serious, but I think they should be totally unrelated.

fhcarter:
unrelated, ignore

[Open JDK Bug](https://bugs.openjdk.java.net/browse/JDK-8266279)

fhcarter:
(Sent email to @rblumer4 in case some other build fails along the way)

PR #276, opened 2021-08-09
Update reconnect protocol to deal with half-open connections

### LINKED ISSUE #275: Half-open connection prevents reconnect
WebSockets have a condition known as _half open_.  This occurs when the connection between two machines is severed (like literally -- network cord unplugged or cut, wifi disabled). When this happens, due to the underlying assumptions built into TCP, the server side (at least) may not notice that the socket connection has been (temporarily) broken.  If the server tries to use things while in the state, it will discover and clean up.  However, when no server activity is ongoing, the server may not know that the client has gone and doesn't clean up appropriately.

Need to support server changes required to make this work.

fhcarter:
Fixes #275 -- will require fix to Vantiq/vantiq-Issues#295 to work completely.

Adds _reconnect secret_ to the connection requests.  This secret allows a connector to usurp its old connection to Vantiq.  This is required when the server fails to notice that the connection between the connector and the server has been unavailable.  This is not uncommon in unstable network connection cases.  When this happens, the connector's attempts to reconnect are rejected by the server because it (thinks it) already has a connection for the source in question.

This fix, in a parallel with the server side fix, allows the connector, using a _reconnect secret_ to assert that it is the connection the server thinks it knows about.  The server will, then, honor the connection & close the old, now defunct, session.

The fix uses parts of the connect extension message not previous used, so old servers will ignore it. And it is optional, so old clients talking to new servers will be fine.

### DIFF
--- a/extjsdk/build.gradle
+++ b/extjsdk/build.gradle
@@ -54,6 +54,10 @@ tasks.withType(Test) {
     if (rootProject.hasProperty("TestVantiqServer")) {
         systemProperty "TestVantiqServer", rootProject.findProperty("TestVantiqServer") ?: "empty"
     }
+
+    if (rootProject.hasProperty("TestRepeatedConnects")) {
+        systemProperty "TestRepeatedConnects", rootProject.findProperty("TestRepeatedConnects") ?: "empty"
+    }
     // Use the build dir as a base to get our various test artifacts.
     systemProperty "buildDir", "${buildDir}"
 }
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
@@ -25,6 +25,7 @@ import java.util.Map;
  * Created by fcarter 7 Jun 2018
  */
 
+@SuppressWarnings("PMD.TooManyFields")
 public class ExtensionServiceMessage {
     public static final String RESOURCE_NAME_SOURCES = "sources";
     public static final String OP_CONFIGURE_EXTENSION = "configureExtension";
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
@@ -43,7 +43,7 @@ public class ExtensionServiceMessage {
     // This is sent on any CONNECT_EXTENSION messages.  It is used when a connector
     // "re-connects" to the server to verify that it's the same connector connecting in
     // (otherwise, the reconnect is rejected).  So, while it's send on all CONNECT... calls,
-    // it's use is in only in the reconnect case (that is, usurping an existing connection).
+    // its use is only in the reconnect case (that is, usurping an existing connection).
     // The name's a bit confusing, but CONNECT_SECRET sounds too much like something you need
     // to connect (which, under normal circumstances, you don't).
     public static final String RECONNECT_SECRET = "reconnectSecret";
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
@@ -39,7 +39,7 @@ public class ExtensionServiceMessage {
     public static final String PROPERTY_MESSAGE_HEADERS = "messageHeaders";
 
     // Set in parameters to allow reconnect after network half-open situation
-    public static String RECONNECT_SECRET = "reconnectSecret";
+    public static final String RECONNECT_SECRET = "reconnectSecret";
 
     public String address;
     public Map    messageHeaders;
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
@@ -38,6 +38,9 @@ public class ExtensionServiceMessage {
     public static final String RESPONSE_ADDRESS_HEADER = "X-Reply-Address";
     public static final String PROPERTY_MESSAGE_HEADERS = "messageHeaders";
 
+    // Set in parameters to allow reconnect after network half-open situation
+    public static String RECONNECT_SECRET = "reconnectSecret";
+
     public String address;
     public Map    messageHeaders;
 
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
@@ -40,6 +40,12 @@ public class ExtensionServiceMessage {
     public static final String PROPERTY_MESSAGE_HEADERS = "messageHeaders";
 
     // Set in parameters to allow reconnect after network half-open situation
+    // This is sent on any CONNECT_EXTENSION messages.  It is used when a connector
+    // "re-connects" to the server to verify that it's the same connector connecting in
+    // (otherwise, the reconnect is rejected).  So, while it's send on all CONNECT... calls,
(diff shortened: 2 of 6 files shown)


extjsdk/build.gradle
```diff
@@ -54,6 +54,10 @@ tasks.withType(Test) {
     if (rootProject.hasProperty("TestVantiqServer")) {
         systemProperty "TestVantiqServer", rootProject.findProperty("TestVantiqServer") ?: "empty"
     }
+
+    if (rootProject.hasProperty("TestRepeatedConnects")) {
+        systemProperty "TestRepeatedConnects", rootProject.findProperty("TestRepeatedConnects") ?: "empty"
+    }
```
fhcarter, line 60: Temporary until we're working with 1.32 systems

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
```diff
@@ -180,6 +181,12 @@ public ExtensionWebSocketClient (String sourceName) {
         this(sourceName, DEFAULT_FAILED_MESSAGE_QUEUE_SIZE);
     }
 
+    /**
+     * Unique identifier for this loader's class.  This is set on load and then left.  This
+     * shared secret allows instances of this client to perform reconnects to the server.
+     */
+    protected final static String clientReconnectSecret = UUID.randomUUID().toString();
+
```
fhcarter, line 189: This is the actual secret.  It is generated at class load/initialization time and allows this connector instance to reconnect at will. It will be the same for all sources to which this connector connects, but that's irrelevant -- we just want to prove ownership.

extjsdk/src/test/java/io/vantiq/extjsdk/RoundTripTestBase.java
```diff
@@ -40,6 +41,8 @@ public static void getProps() {
         testSourceName = System.getProperty("EntConTestSourceName", "testSourceName");
         testTypeName = System.getProperty("EntConTestTypeName", "testTypeName");
         testRuleName = System.getProperty("EntConTestRuleName", "testRuleName");
+        testRepeatedConnectsEnabled =
+                Boolean.valueOf(System.getProperty("TestRepeatedConnects", "false"));
```
fhcarter, line 45: Temporary -- we can remove when we're testing mostly against 1.32+ systems

extjsdk/src/test/java/io/vantiq/extjsdk/TestWithServer.java
```diff
@@ -26,6 +27,39 @@ public static void cleanup() {
...
+        // Now, to verify that things work as expected, repeatedly connect w/o close
+        // This is verifying that the reconnect secret is being passed & used correctly.
+
+        for (int i = 0; i < repCount; i++ ) {
+            client = new ExtensionWebSocketClient(SOURCE_NAME);
+            client.initiateFullConnection(testVantiqServer, testAuthToken).get();
+            assertTrue("Failed to open connection", client.isOpen());
+            assertTrue("Failed to authenticate", client.isAuthed());
+            assertTrue("Failed to connect to source (may fail if run against Vantiq version < 1.32)",
+                    client.isConnected());
```
fhcarter, line 59: but just in case, we'll alert...

fhcarter:
Related Vantiq PR: Vantiq/ag2rs#8032

fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/28/)

fhcarter:
@rblumer4, this may be related to the OPC errors.  Looks like it all stems from some bug(s) in OpenJDK...

https://github.com/bcgit/bc-java/issues/941

extjsdk/src/test/resources/simulateUnstableNetwork.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# Shell script for Mac OS that will turn off then on wifi to monkey with network.
+# This allows for manual testing for connector disconnection recovery.
+
+for ((i = 0; i < 100; i++)) do
```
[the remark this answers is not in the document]
fhcarter: This should be ignored as per spellcheck directive atop file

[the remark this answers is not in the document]
fhcarter: This should be ignored as per spellcheck directive atop file

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
```diff
@@ -38,6 +39,9 @@
     public static final String RESPONSE_ADDRESS_HEADER = "X-Reply-Address";
     public static final String PROPERTY_MESSAGE_HEADERS = "messageHeaders";
 
+    // Set in parameters to allow reconnect after network half-open situation
+    public static final String RECONNECT_SECRET = "reconnectSecret";
```
namirfa, line 43: Should this be the "reconnect secret" or "connect secret"?

I think it's fine either way, but the secret would be used for both the first connection as well as any other reconnection afterwards.

extjsdk/src/test/java/io/vantiq/extjsdk/RoundTripTestBase.java
namirfa: We should probably add an issue reminding us to remove this once 1.32 is out.

namirfa (approved):
LGTM

Two minor comments, one about adding a new issue and one about a potential name change. Feel free to ignore.

fhcarter:
Issue to remove stuff is #277

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
fhcarter: I'd prefer to leave it as is.  CONNECT_SECRET sounds like something that you need to connect (and, generally, you don't). It's also invisible to users of the SDK.

I'll think if there's a comment I can add that clarifies things.

namirfa: Works for me!

fhcarter:
[Jenkins re-run after all the changes...](extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java)

extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java
```diff
@@ -40,6 +40,12 @@
     public static final String PROPERTY_MESSAGE_HEADERS = "messageHeaders";
 
     // Set in parameters to allow reconnect after network half-open situation
+    // This is sent on any CONNECT_EXTENSION messages.  It is used when a connector
+    // "re-connects" to the server to verify that it's the same connector connecting in
+    // (otherwise, the reconnect is rejected).  So, while it's send on all CONNECT... calls,
+    // it's use is in only in the reconnect case (that is, usurping an existing connection).
+    // The name's a bit confusing, but CONNECT_SECRET sounds too much like something you need
+    // to connect (which, under normal circumstances, you don't).
```
namirfa, line 48:
```suggestion
    // This is sent on any CONNECT_EXTENSION messages.  It is used when a connector
    // "re-connects" to the server to verify that it's the same connector connecting in
    // (otherwise, the reconnect is rejected).  So, while it's sent on all CONNECT... calls,
    // its use is only in the reconnect case (that is, usurping an existing connection).
    // The name's a bit confusing, but CONNECT_SECRET sounds too much like something you need
    // to connect (which, under normal circumstances, you don't).
```

namirfa:
> [Jenkins re-run after all the changes...](extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionServiceMessage.java)
> 
> (no changes)

(wrong link)

fhcarter:
Note: This to fully work, this requires an update to the Vantiq server.  This is anticipated in release 1.32.

PR #288, opened 2021-11-11
Add documentation for baseConfigProperties and include examples

### LINKED ISSUE #286: Document baseConfigProperties and change example sourceimpls

namirfa:
Closes #286 

Adding documentation for baseConfigProperties, along with examples for each connector's sourceimpl document. Also moved UDP sourceimpl from `main/resources` to `test/resources`

### DIFF
--- a/objectRecognitionSource/src/test/resources/objRecImpl.json
+++ b/objectRecognitionSource/src/test/resources/objRecImpl.json
@@ -2,5 +2,9 @@
    "name" : "OBJECT_RECOGNITION",
    "baseType" : "EXTENSION",
    "verticle" : "service:extensionSource",
-   "config" : {}
+   "config" : {},
+   "baseConfigProperties": [
+      "objRecConfig.dataSource.type",
+      "objRecConfig.neuralNet.type"
+   ]
 }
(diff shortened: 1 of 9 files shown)


objectRecognitionSource/src/test/resources/objRecImpl.json
```diff
@@ -2,5 +2,9 @@
    "name" : "OBJECT_RECOGNITION",
    "baseType" : "EXTENSION",
    "verticle" : "service:extensionSource",
-   "config" : {}
+   "config" : {},
+   "baseConfigProperties": [
+      "objRecConfig.dataSource.type",
+      "objRecConfig.neuralNet.type"
```
fhcarter, line 8: wouldn't we typically want the camera address here, too? Or is that too tricky since it's dependent upon the datasource type?

fhcarter (approved):
LGTM

objectRecognitionSource/src/test/resources/objRecImpl.json
namirfa: Yep- because different dataSource types offer different config options, I wasn't sure if we should add the `camera` or not. Now that I think of it, I'm pretty sure all of them require a `camera`, so I think I can add that without issue.

PR #291, opened 2021-12-10
Correctly handle simultaneous queries involving dates

### LINKED ISSUE #290: JDBC: Multiple, simultaneous queries including dates may return spurious results
When many queries are run at the same time & those queries return dates, these dates can get mixed during the return process.

fhcarter:
Fixes #290 

When there were multiple threads (_e.g._, when different Vantiq events caused queries to run simultaneously) were running queries against tables containing dates (including cases where it was the same table), the date strings returned could be garbled or confused.  This was due to the use of Java's `SimpleDateFormat` in a multi-threaded manner. This class is not thread safe, so...

Moved date formatters into method that does the database `SELECT` statement;  consequently, each thread gets its own copy that it can serially reuse.

Added test that runs 150 queries simultaneously -- w/o fix, things died with only a few threads. With fix, the 150 threads run fine. I ramped it to 1500 without issue, but there's no reason to keep that level in the tests.

### DIFF
--- a/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
+++ b/jdbcSource/src/main/java/io/vantiq/extsrc/jdbcSource/JDBC.java
@@ -22,6 +22,7 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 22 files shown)


fhcarter:
wrong stuff crept in...

PR #303, opened 2021-12-17
Fix error from extjsdk:fatJar task

### LINKED ISSUE #302: Extjsdk: task fatJar causes duplicate file issue re: licenses
Function of gradle upgrade.

fhcarter:
Fixes #302

Add `duplicatesStrategy` to `fatJar` task.

Numerous things include META-INF/LICENSE stuff & we don't want to fail as a result...

### DIFF
--- a/extjsdk/build.gradle
+++ b/extjsdk/build.gradle
@@ -93,6 +93,7 @@ dependencies {
 task fatJar(type: Jar) {
     appendix = 'fat'
     from sourceSets.main.output
+    duplicatesStrategy 'include'
     from(".") {
         include "README.md"
         include "LICENSE/*"


extjsdk/build.gradle
```diff
@@ -93,6 +93,7 @@ dependencies {
 task fatJar(type: Jar) {
     appendix = 'fat'
     from sourceSets.main.output
+    duplicatesStrategy 'include'
```
fhcarter, line 96: Gradle 7.x gives an error by default on duplicate files in any copy-derivative task.  Have to specify how to handle duplicates.  We choose to allow them to override one another...

AlexBlumer (approved):
LGTM

PR #325, opened 2022-05-19
Use connector SDK probe capabilities

### LINKED ISSUE #314: Python extpsdk & pythonExecSource -- generate setup.cfg from templatized version
Also, add how-to release documents to each.

### LINKED ISSUE #316: PythonExecSource: Use probes created in the Extpsdk
DependsOn #315

fhcarter:
Fixes #316 
Fixes #314 

Re: 316

When connector set starts up (iff running in K8s), declare healthy.  After that, it's up to the connector itself to declare itself unhealthy.  In this connector, we do that only when we've been told about an unrecoverable memory error (_i.e._, out of...).  However, that user can declare things unhealthy, and then things get restarted as expected.  This is the hand testing done to verify things.

Re: 314:

The Python generate process can have a number of different files that have embedded version numbers. This is problematic when you have to remember to change them multiple places.

This PR changes the build process so that the version number to be used for this generation is set in the build.gradle file & that's used to generate (here) the setup.cfg file used by the python distribution & publishing mechanisms.

Note that the version numbers are in sort-of non-standard properties so that the same properties can be used throughout the other Vantiq Python products.


pythonExecSource/src/main/python/pyExecConnector.py
```diff
@@ -673,6 +681,11 @@ async def run(self):
             pec.establish_handlers()
 
         self.logger.info('Running PyExecConnector.')
```
fhcarter, line 683: Look for env variable set up by K8s to determine whether to start the TCP probe handler

pythonExecSource/src/main/python/pyExecConnector.py
```diff
@@ -659,6 +666,7 @@ class Connectors:
     def __init__(self):
         self.connector_set = VantiqConnectorSet()
         self.logger = logging.getLogger('Vantiq.PyExecConnector')
+        self.logger.setLevel(logging.DEBUG)
```
AlexBlumer, line 669: Is this just debugging or supposed to be committed?

AlexBlumer (approved):
LGTM

pythonExecSource/src/main/python/pyExecConnector.py
fhcarter: It's early enough in the process (and, if used at all, used by our folks) that I'd prefer to get more data now.  I'll mark it as a TODO & remove it after things get burned in a bit.

PR #327, opened 2022-05-27
Fix python exec source docs to use correct config doc format.  Also, update versions & better examples.

### LINKED ISSUE #326: PythonExecSource:  README: server.config/source should be server.config/sources

fhcarter:
Fixes #326 

Fix refs to `sources` in config docs.

Also, add some python example code to the documents & fix some other typos.

Update to use more recent _vantiqsdk_ & bump our version accordingly.

### DIFF
--- a/pythonExecSource/README.md
+++ b/pythonExecSource/README.md
@@ -19,6 +19,8 @@ The Vantiq Python Execution Connector requires Python version 3.8 or better.
 
 ## Quick Start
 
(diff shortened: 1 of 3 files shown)


fhcarter:
Addresses some things found by @skyler1ackerman during his experiments.  Holding this for a bit pending other things found by the first non-me to use it.  I'm sure there will be more.

PR #329, opened 2022-06-01
Handle occasional status message

### LINKED ISSUE #328: extpsdk: server's occasional status message confuses SDK
Vantiq server _sometimes_ sends a status message after a connect, sometimes not.  It appears to be more prevalent with https connections, but not exclusive.

Add code so that when a status message is received there (and is successful), we read again to get the connect extension message.  Otherwise, the protocol is out of sync and the connection never completes.

fhcarter:
Fixes #328

Vantiq server _sometimes_ sends a status message after a connect, sometimes not.  It appears to be more prevalent with https connections, but not exclusive.

Add code so that when a status message is received there (and is successful), we read again to get the connect extension message.  Otherwise, the protocol is out of sync and the connection never completes.

### DIFF
--- a/extpsdk/src/main/python/vantiqconnectorsdk.py
+++ b/extpsdk/src/main/python/vantiqconnectorsdk.py
@@ -434,6 +434,9 @@ class VantiqSourceConnection:
                     status = resp[_STATUS]
                     if status >= 300:
                         return _CONNECTION_FAILED
+                    # Sometimes we get a status message back.  If so & it's OK, wait for the config message
+                    raw_resp = await websocket.recv()
+                    resp = json.loads(raw_resp)
 
                 # Otherwise, we should have a configExtension message
                 if _OPERATION in resp.keys():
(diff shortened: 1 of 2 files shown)


extpsdk/src/main/python/vantiqconnectorsdk.py
```diff
@@ -434,6 +434,9 @@ async def _perform_connection(self, do_pings: bool) -> string:
                     status = resp[_STATUS]
                     if status >= 300:
                         return _CONNECTION_FAILED
+                    # Sometimes we get a status message back.  If so & it's OK, wait for the config message
+                    raw_resp = await websocket.recv()
+                    resp = json.loads(raw_resp)
```
AlexBlumer, line 439: This should probably be part of a loop. Otherwise, it looks like we don't handle a failure or second status message after the first status message.

AlexBlumer (approved):
LGTM

PR #346, opened 2022-08-15
Enable including of Base64-encoded image in results

### LINKED ISSUE #317: ObjRec: Add option to include the image as a base64 string content to the payload instead of file
Would like an option to include the image file content in the payload itself as a base64 encoded string which can be used directory in HTML img tags ". 

@bruden01 created his own one-off that does this but it would be nice to have a flag we could set with the image options where instead of vantiq, local, both we can also do inline or something like that to put the image content into the payload.

fhcarter:
Fixes #317 

Adds `includeEncodedImage` to configuration & query parameters.  When set, the results (for a query, when `sendFullResponse` is true) will include a property `encodedImage` whose value is the Base64 encoded image on which the results are based.  If `labelImage` is true, the labeling will be included in the resulting image.

Added tests, etc., for same.

### DIFF
--- a/jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBCBase.java
+++ b/jdbcSource/src/test/java/io/vantiq/extsrc/jdbcSource/TestJDBCBase.java
@@ -93,7 +93,6 @@ public class TestJDBCBase {
     protected static void deleteSourceImpl(Vantiq vantiq) {
         if (createdImpl) {
             vantiq.deleteOne(VANTIQ_SOURCE_IMPL, JDBC_SRC_TYPE);
-            createdImpl = false;
         }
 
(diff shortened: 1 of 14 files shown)


fhcarter:
_ReadMe_ updates to come.  Wanted to get sider started...

jcguion (approved):
LGTM

fhcarter:
(None of the disabling for check style things I've tried worked.  So just ignoring them for now.)

AlexBlumer (approved):
LGTM

fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/89/)

fhcarter:
(Complete run against Vantiq run locally)

PR #355, opened 2022-12-21
Fix sporadic reconnect/reconfig issue

### LINKED ISSUE #354: ExtJSdk: RECONNECT_REQUIRED/client.doCoreReconnect() sometimes fails with error from server
Server failure of the form "...because the source ... does not have an associated stream..."

The issue here is that there is a window in the server where a configuration change causes the source to restart (and possibly migrate to a different cluster member).  During that window, the source is unavailable.

Solution is to add retry to the underlying SDK.

fhcarter:
Fixes #354 

Adds retry logic to ExtJSdk's ExtensionWebSocketClient.doCoreReconnect().  Due to the way the server must function, there is a small window where a restarting source can be in a discombobulated state.  When we encounter this, retry the connect a few times.  This should close the window where errors of the for OP_CONNECT_EXTENSION failed because there was no associates stream service."  These were rare, but did occur.

Also fixed issue with date formatting test.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -134,6 +134,9 @@ public class ExtensionWebSocketClient {
      * Whether it should automatically send a connection message after receiving a reconnect message
      */
     boolean autoReconnect = false;
+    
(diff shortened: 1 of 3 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/97/) in progress.

Local run with server to fully test things.

jcguion (approved):
LGTM

PR #360, opened 2023-01-11
Add handling for remote (or other unexpected) close of web socket.

### LINKED ISSUE #359: ExtJSdk: Websocket in inconsistent state is ignored
When web socket connection to vantiq is in an inconsistent state (remote closed, lost connection, lost intermediary, etc.), that fact is not reported by a send.  When the send is done, the exception is trapped and logged, but not reported the program.

fhcarter:
Fixes #359 

When a send() request fails with an illegalStateException, it means that the websocket is no longer serviceable.  This may be due to the server (remote) closing the socket or some other issue in the network connection.  When we detect this, we had previously logged the error but otherwise done nothing.

Change this behavior so that the client will close the connection and throw an exception.  The closing will cause the close handler to be called (assuming one has been configured), allowing the client to recover as is appropriate.  Alternatively,, the client can trap the exception and react to that.

Added emulation & test for this situation.

### DIFF
--- a/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
+++ b/extjsdk/src/main/java/io/vantiq/extjsdk/ExtensionWebSocketClient.java
@@ -586,8 +586,12 @@ public class ExtensionWebSocketClient {
                     this.webSocket.send(ByteString.of(bytes));
                 }
             }
-        }
-        catch (Exception e) {
+        } catch (IllegalStateException ise) {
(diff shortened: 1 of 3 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/104/)

fhcarter:
@msilva123

AlexBlumer (approved):
LGTM

msilva123:
Awesome, thanks!

PR #375, opened 2023-04-01
Add warning when createConsumer() called multiple times.

fhcarter:
This may be due to caller error -- there could be > 1 route defined using the same consumer (which is an error).  However, it is also the case that defining (even a single) route with no id can cause this to happen (as is the case in this issue). This appears to be an issue with Apache camel itself.  See Camel issues [14969](https://issues.apache.org/jira/browse/CAMEL-14969) and [13651](https://issues.apache.org/jira/browse/CAMEL-13651) which appear to be similar but not quite the same.

To help folks resolve this, we'll dump out an error and a hint.  Tried to return the same consumer, but that doesn't avoid the issue.


jcguion (approved):
LGTM

PR #376, opened 2023-04-17
Add collected classPath to main jar file's manifest.

### LINKED ISSUE #371: camelConnector build creates a seemingly invalid windows bat file that does not run
After doing a clean gradle build of the camelConnector I cannot run the windows bat file that is built. I can open an Ubuntu shell and run the shell script no problem, I can also just run the java command manually and that works, I suspect there is a typo in the file somewhere (see attached). 

[camelConnector.txt](https://github.com/Vantiq/vantiq-extension-sources/files/11092896/camelConnector.txt)

Here is the error: 

> \>  camelConnector.bat  server.config
The input line is too long.
The syntax of the command is incorrect.

fhcarter:
Fixes #371

Add collected classPath to main jar file's manifest.  Then, set class path (for application plugin's `startScript` to contain only the main jar file.  The consequence of this is that the class path is listed in the manifest of the camelConnector.jar file.

This allows us to avoid expanding the classPath in the script file.  The potentially long list of things here causes issues on Windows with the cmd.exe file having limits on the command line length (and, potentially, the variable content length though that's less of an issue).

Also set the main class name which allows the connector to be run by just using a `java -jar` command.  Using that will be for a later fix, if ever.

### DIFF
--- a/camelConnector/build.gradle
+++ b/camelConnector/build.gradle
@@ -12,6 +12,8 @@ plugins {
 mainClassName = 'io.vantiq.extsrc.camelconn.connector.CamelMain'
 
 startScripts{
+    // Don't neeed all the class path stuff here since we've injected it into th main jar's manifest
+    classpath = jar.outputs.files
     doLast{
         def windowsScriptFile = file getWindowsScript()
         def unixScriptFile = file getUnixScript()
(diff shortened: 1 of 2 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/)

Also ran locally using salesforce & another bug-fix configuration & all worked as expected.

AlexBlumer (approved):
LGTM

fhcarter:
Hand inspected the `camelConnector.bat` file produced, and the previously offending line (as per bug report) is now 

```
set CLASSPATH=%APP_HOME%\logConfig;%APP_HOME%\lib\camelConnector.jar
```

which should be well under the 8K limit :-)

camelConnector/build.gradle
```diff
@@ -12,6 +12,8 @@ plugins {
 mainClassName = 'io.vantiq.extsrc.camelconn.connector.CamelMain'
 
 startScripts{
+    // Don't neeed all the class path stuff here since we've injected it into th main jar's manifest
```
rblumer4, line 15: typo:  e migrated from "the" ...
```suggestion
    // Don't need all the class path stuff here since we've injected it into the main jar's manifest
```

PR #378, opened 2023-05-17
ExtJSdk: extension source stops retrying to connect if source is momentarily missing 

### LINKED ISSUE #377: ExtJSdk: extension source stops retrying to connect if source is momentarily missing
If the source definition is momentarily removed from the Vantiq server to be added back some time later, the extension source stops retrying to reconnect. This requires to manually restart the extension source.

jcguion:
If we cannot automatically reconnect to the source (and failed enough trying), close the connection so the extension source code can decide what to do about it.

Testing this scenario was based on manual testing. In this case, removing a source definition while the extension source was connected, observing failure, recreating the source definition and validating that the extension source could reconnect automatically.

Close #377


fhcarter (approved):
LGTM

PR #405, opened 2023-09-15
Cleanup/testing enhancement: Include all possible modules regardless of driver presence

### LINKED ISSUE #404: Testing: Improve testing by including all modules possible
Make ones with specialized drivers dependent on those drivers only for running.  Some simple tests and compilation, etc., can be done without the drivers.

fhcarter:
Fixes #404 

Allows for a bit more verification when used in generalized testing scenarios.  Previously, we had ignored entire modules if the associated driver definitions (JDBC, JMS) were not present.  This PR changes that so that we'll include the module, build it, and do whatever testing is possible (often not much, but some isolated unit tests are feasible).  This may allow us to detect broken things earlier.

At present, this affects only the JDBC & JMS sources -- as these are based on general interfaces at compile tile.  The EasyModBus source requires the jar file for compilation of the sources, so without the "driver" file, this can't proceed.

### DIFF
--- a/jdbcSource/build.gradle
+++ b/jdbcSource/build.gradle
@@ -66,7 +66,9 @@ dependencies {
     implementation "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}"
     
     //  Compiling relevant JDBC Driver .jar file
-    runtimeOnly files("${System.env.JDBC_DRIVER_LOC}")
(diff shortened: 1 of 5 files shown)


fhcarter:
Clean [Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/148/)

jcguion (approved):
LGTM

PR #407, opened 2023-09-20
Update camel version & use the version in package names

fhcarter:
Another in the series merging into a staging branch.

Update to Camel version 3.21.0. This is an LTS version (Long Term Support) and may be the last major major update in the version 3 series.  Version 4 has been released, but we'll deal with that at a later date.  We update to a version that includes kamelets, but we need to keep the kamelet version & connector Camel versions in sync since various config parameters may be adapted along the way.

Using Camel version in package names allows us  to publish multiple versions should that be necessary, and allows more than one version to coexist if necessary. There are differences between the versions that show up at the assembly level -- typically in configuration parameters.

Fix some issues in configuration properties.  One, the lack of enums, still exists -- we currently generate what we speculate will be supported, but that's subject to change so we may pull it before publication of assemblies.

Include descriptive (usually use) information in an overview document.  This information is extracted from the kamelet definition used to generate these assemblies


buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -501,7 +515,16 @@ class AssemblyResourceGeneration extends DefaultTask {
             log.debug("\t processing property {}: {}", pName, pDesc)
 
             // Properties are required if there is no default...
-            Map<String, Object> propDesc = [ required: pDesc.default == null ]
```
fhcarter, line 504: simplistic setting of required (above) was incorrect. Use `required` property as provided (this may be new).

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -523,14 +546,18 @@ class AssemblyResourceGeneration extends DefaultTask {
                 ((String) pDesc[PROPERTY_X_DESCRIPTORS])?.contains(DISPLAY_HIDE_VALUE)) {
                 // for things that are passwords or are marked to be masked in display, we will require a secret.
                 propDesc.type = 'Secret'
-                propDesc.description += ' (Please provide the name of the Vantiq Secret containing ' + 'this value.)'
+                propDesc.description += ' (Please provide the name of the Vantiq Secret containing this value.)'
             }
             if (pDesc.enum != null) {
-                propDesc.enum = pDesc.enum
```
fhcarter, line 529: Re: the section below.  This speculates on the interface that may work.  Whether this survives until publication or not remains to be seen.  But it works as is -- if this interface is adopted, then we'll immediately take advantage of it.

camelComponent/src/main/java/io/vantiq/extsrc/camel/VantiqConsumer.java
```diff
@@ -47,7 +46,7 @@ public VantiqConsumer(VantiqEndpoint endpoint, Processor processor) {
     protected void doStart() throws Exception {
         super.doStart();
         endpoint.startup();
-        vantiqClient = endpoint.getVantiqClient();
```
fhcarter, line 50: fix extraneous problem when reconfig is required.

camelConnector/src/main/resources/artifactMap.json
```diff
@@ -1,5 +1,5 @@
 {
-    "camelVersion": "3.18.3",
```
fhcarter, line 2: side effect of version update.  But we don't really want to run this for each build, so we check it in...

fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/151/)

jcguion (approved):
LGTM

PR #410, opened 2023-09-22
Camel Component: Use correct mechanism to output the JSON nodes

### LINKED ISSUE #408: Camel Component: Json output puts strings in quotes

fhcarter:
Fixes #408

W/o fix, json output of strings was "some string" rather than `some string` which was incorrect.

Also fixes issue with `vantiq://server.config` style cases being closed when they shouldn't.  In that case, the client is under control of the Camel connector & the closing thereof should be left to said connector.


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/152/)

jcguion (approved):
LGTM

PR #411, opened 2023-09-22
Camel Connector: Add support for wrapping prop values in RAW().

### LINKED ISSUE #409: Camel Connector: Allow properties to require RAW handling
When providing RAW values, it is sometime necessary to wrap them in RAW indicators -- this tells the Camel mechanism to alter any URL encoding going on.  For some cases, we'd like to have the connector mechanism handle this, so add handling for a new config property which is a set of lists of property names that require RAW handling.  Upon receipt of the configuration, these values will be wrapped in RAW indicators before being given to Camel's Property component.

fhcarter:
Adds config parameter which is a set of named lists for property names whose values should be wrapped with RAW() (or RAW{}).  This allows users configuring things to skip this step of the Camel connector is so configured.

Fixes #409

### DIFF
--- a/camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
+++ b/camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
@@ -254,9 +254,11 @@ public class CamelHandleConfiguration extends Handler<ExtensionServiceMessage> {
             Map<String, List<String>> raw = (Map<String, List<String>>) camelConfig.get(RAW_REQUIRED);
             List<String> rawRequired = new ArrayList<>();
             // Create a list from the various categories that may be in this set...
-            raw.forEach( (k, v) -> {
-                rawRequired.addAll(v);
-            });
+            if (raw != null) {
+                raw.forEach((k, v) -> {
+                    rawRequired.addAll(v);
+                });
+            }
             // Camel wants these as a Java Properties object, so perform the conversion as required.
             Properties propVals = null;
             if (input != null) {
--- a/camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
+++ b/camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
@@ -61,9 +61,11 @@ public class CamelHandleConfiguration extends Handler<ExtensionServiceMessage> {
     public static final String ROUTES_FORMAT = "routesFormat";
     public static final String COMPONENT_PROPERTIES = "componentProperties";
     public static final String PROPERTY_VALUES = "propertyValues";
+    public static final String RAW_REQUIRED = "rawValuesRequired";
+    public static final String DISCOVERED_RAW = "discovered";
+    public static final String CONFIGGURED_RAW = "configured";
     
     public static final String VANTIQ = "vantiq";
-    
     public static final String GENERAL = "general";
     public static final String COMPONENT_CACHE = "componentCacheDirectory";
     public static final String COMPONENT_CACHE_DEFAULT = "componentCache";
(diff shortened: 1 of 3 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/153/)

camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
```diff
@@ -61,9 +61,11 @@ public class CamelHandleConfiguration extends Handler<ExtensionServiceMessage> {
     public static final String ROUTES_FORMAT = "routesFormat";
     public static final String COMPONENT_PROPERTIES = "componentProperties";
     public static final String PROPERTY_VALUES = "propertyValues";
+    public static final String RAW_REQUIRED = "rawValuesRequired";
+    public static final String DISCOVERED_RAW = "discovered";
+    public static final String CONFIGGURED_RAW = "configured";
```
AlexBlumer, line 66:
```suggestion
    public static final String CONFIGURED_RAW = "configured";
```

camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
```diff
@@ -258,12 +250,43 @@ public boolean runCamelApp(Map<String, Object> camelConfig, Map<String, Object>
...
+                                String endRaw = ")";
+                                if (val.contains(")")) {
+                                    if (!val.contains("}")) {
+                                        startRaw = "RAW{";
+                                        endRaw = "}";
+                                    }
+                                }
+                                propVals.setProperty(p.getKey(), startRaw + val + endRaw);
+                            } else {
+                                // If we're suppressing or we havwe a value with both ) & }, leave things alone
```
AlexBlumer, line 283:
```suggestion
                                // If we're suppressing or we have a value with both ) & }, leave things alone
```

AlexBlumer: Why does it not need to wrap if the string has both a `)` and a `}`? If it's just that it would mess with the RAW parsing, that seems like it should cause an error rather than just ignore it.

AlexBlumer (approved):
LGTM, with one question

camelConnector/src/main/java/io/vantiq/extsrc/camelconn/connector/CamelHandleConfiguration.java
fhcarter: I think that the receiver would see an error (since the property value isn’t wrapped), but I don’t know that for certain. So, rather than our refusing to do the work, it seemed better to pass it thru and let the underlying systems sort things out. In this case, it is our generation that decided that things likely needed to be wrapped; it may be that that’s not required — it may be that the underlying systems can sort it out. Since this is 5th party integration atop 4th party atop 3rd party, much of it open source, it seemed better here to just leave things in their natural state and let the _nature_ take its course.

fhcarter: (Sorry — didn’t see this remark yesterday, so I’m a bit late with the response.)

fhcarter: Updated code -- streamlined a bit and added comments about what & why.  Also added ability to bypass auto wrapping.  Added tests

fhcarter:
[Jenkins run redux](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/158/)

AlexBlumer (approved):
LGTM

PR #412, opened 2023-09-22
Improve handling of required properties & RAW wrapping of property values

fhcarter:
Use spec/definition/required as primary indicator, but allow properties to have a `required` property that is the ultimate truth.  Ensure that this value is set in the assembly definition so that assembly installation requires specification of values here.

As part of assembly generation, track the properties that are marked as passwords, credentials, or those that have UI characteristics that say they shouldn't be displayed. For these,

* Add to their description suggesting that they be specfied as Vantiq Secrets and have their value entered as @secrets(<<secretName>>).
* Mark them in the generated source as needing Camel RAW wrapping.

Corresponding fixes (#409/#411) will have the Camel connector appropriately wrapping the property values for the Properties component.  These two actions, in concert, allow the assembly installer to remain unaware of the RAW() requirements.  It is still available (the user can add RAW(some value) as the assembly config property value, but for many cases, it won't be necessary.

Also, clean up internal rules to preserve the headers, and generate the internal route with the json generation specified.

No issue entered.


jcguion (approved):
LGTM

PR #413, opened 2023-09-28
Camel Component: time and byte handling

fhcarter:
Corrects issue noticed while testing against aws s3.

When byte stream is given to the component, it fails -- not knowing what to do.  It's been enhanced to check the byte stream -- if it's convertible to a map, we'll do that.  If it's a valid String, we'll convert to that, other wise we leave it as byte array.  In the cases of a byte array or String, we send back a message that's a Map with one key -- byteVal or stringVal.

Over time, we're likely to try, in some circumstances, to make note of what's going on & upload a document or whatever.  However, the component itself doesn't really have that knowledge, so that'll be a longer-term plan.

Also -- dates, instants, and their ilk are not currently passed up very well.   So, we now convert any of these to their equivalent strings.

Added associated tests.


AlexBlumer (approved):
LGTM

PR #415, opened 2023-09-28
Generate route templates rather than routes

fhcarter:
This is what's used in kamelet land, so best to keep with it.  Corresponding change to camelConnector enables these changes.

Part of this was learning from the declared dependencies in the kamelet files. The information there, while not necessarily complete, allows us to pick up dependencies that might be required for beans embedded in the route templates.

This change caused a change in behavior, at least for the aws kamelets. Specifically, without including the declared dependencies, the AWS endpoints were not RAW() wrapping their endpoint credentials.  With the dependency inclusion, things are now wrapped. So, for now, I've disabled auto wrapping. It may prove necessary for other kamelets, so the code is there, albeit commented out. If we need to  add it back conditionally, we can.

This situation is curious in that the camel aws* components _should_ include whatever dependencies they need in the their respective pom.xml files, so that dependency management can take care of such things. But it appears not to, so handling this changed things.

### DIFF
--- a/buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
+++ b/buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
@@ -41,7 +41,8 @@ import java.util.regex.Pattern
 class AssemblyResourceGeneration extends DefaultTask {
     public static final String VANTIQ_SERVER_CONFIG = 'vantiq://server.config'
     public static final String VANTIQ_SERVER_CONFIG_JSON = VANTIQ_SERVER_CONFIG + '?consumerOutputJson=true'
-    public static final String VANTIQ_SERVER_CONFIG_STRUCTURED = VANTIQ_SERVER_CONFIG + '?consumerOutputJson=true' +
+    public static final String VANTIQ_SERVER_CONFIG_STRUCTURED = VANTIQ_SERVER_CONFIG + '?structuredMessageHeader=true'
+    public static final String VANTIQ_SERVER_CONFIG_JSON_STRUCTURED = VANTIQ_SERVER_CONFIG_JSON +
         '&structuredMessageHeader=true'
 
     public static final String KAMELETS_RESOURCE_FOLDER = 'kamelets'
@@ -70,7 +71,7 @@ class AssemblyResourceGeneration extends DefaultTask {
 
     public static final String FROM_ROUTE_TAG = 'from'
     public static final String ID_ROUTE_TAG = 'id'
-    public static final String ROUTE_ROUTE_TAG = 'route'
+    public static final String ROUTE_TEMPLATE_TAG = 'route-template'
     public static final String TO_ROUTE_TAG = 'to'
     public static final String URI_ROUTE_TAG = 'uri'
 
@@ -94,6 +95,7 @@ class AssemblyResourceGeneration extends DefaultTask {
     public static final String CONFIG_CAMEL_GENERAL = 'general'
     public static final String CAMEL_RUNTIME_APPNAME = 'appName'
     public static final String CAMEL_RUNTIME_PROPERTY_VALUES = 'propertyValues'
+    public static final String GENERAL_ADDITIONAL_LIBRARIES = 'additionalLibraries'
     public static final String CAMEL_RAW_HANDLING = 'rawValuesRequired'
     // TODO: Add ability to add some overrides here where we want Camel RAW handling to be performed
     public static final String CAMEL_DISCOVERED_RAW = 'discovered'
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -251,6 +253,13 @@ class AssemblyResourceGeneration extends DefaultTask {
         Map<String, Object> yamlMap = getYamlEnt(yamlLoader, classLoader, kameletDefinition)
         log.debug('    \n<<<Result: parsed.size(): {}, parsed: {}>>>\n', yamlMap.size(),
             yamlMap.get('spec'))
```
fhcarter, line 255: check labels as defined to make sure they match the naming conventions we're using.

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -261,7 +270,21 @@ class AssemblyResourceGeneration extends DefaultTask {
             .get('definition')).get('description')
         List<String> requiredProps = (List<String>) ((Map<String, Object>)((Map<String, Object>) yamlMap.get('spec'))
             .get('definition')).get('required')
-        log.debug('Required Props from kamelet defininition: {}', requiredProps ?: '<<empty>>')
+        log.debug('Required Props from kamelet definition: {}', requiredProps ?: '<<empty>>')
+        List<String> dependencies = (List<String>) ((Map<String, Object>) yamlMap.get('spec')).get('dependencies')
+        // Dependencies can also be declared at the type level, so grab those as well.
```
fhcarter, line 275: grab the declared dependencies from the top level and on a per-type basis

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -326,7 +349,7 @@ class AssemblyResourceGeneration extends DefaultTask {
         Map routeDoc = writeRoutesDocument(assemblyRoot, packageName, kamName, routesDocumentString)
         Map overviewDoc = (overview ? writeOverviewDocument(assemblyRoot, packageName, kamName, overview) : null)
         Map sourceDef = addSourceDefinition(assemblyRoot, packageName, kamName,
-            routeDoc.path.fileName.toString(), props)
```
fhcarter, line 329: these are added to the source def'n as `additionalLibraries`, a feature that was already there (and worked!).

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -455,13 +480,18 @@ class AssemblyResourceGeneration extends DefaultTask {
...
             if (aProp.value instanceof Map) {
                 Map pDesc = aProp.value as Map
-                if (PASSWORD_LIKE.contains(pDesc.format?.toLowerCase()) ||
-                    (pDesc[PROPERTY_X_DESCRIPTORS] != null &&
-                        DISPLAY_HIDE_VALUES.contains(((String) pDesc[PROPERTY_X_DESCRIPTORS])))) {
-                    // for things that are passwords or are marked to be masked in display, tell connector to instruct
-                    // Camel to use RAW handling
-                    rawHandlingRequired << aProp.key
-                }
+                // FIXME: With more full dependency handling, this no longer seems necessary, at least for aws things.  Not
```
fhcarter, line 483: leaving this here as I suspect it returns on a conditional, per-kamelet, basis

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -802,8 +847,8 @@ class AssemblyResourceGeneration extends DefaultTask {
             if (k == FROM_ROUTE_TAG && stepDef != null) {
                 if (stepDef.containsKey(URI_ROUTE_TAG) && stepDef.get(URI_ROUTE_TAG) instanceof String &&
                         ((String) stepDef.get(URI_ROUTE_TAG)) == KAMELET_SOURCE) {
-                    stepDef.put(URI_ROUTE_TAG, VANTIQ_SERVER_CONFIG_STRUCTURED)
-                    log.debug('Replaced {} with{}', stepDef.get('uri'), VANTIQ_SERVER_CONFIG_STRUCTURED)
```
fhcarter, line 806: generate all messages in { headers: ..., message: ... } vail object format.

buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
```diff
@@ -906,7 +951,7 @@ RULE ${ruleName}
 WHEN EVENT OCCURS ON "/sources/${sourceName}" as svcMsg
 
 log.debug("Source ${sourceName} rule got message {} to forward to service  {}", [ svcMsg, 
-    "${packageName + '.' + serviceName}/${serviceName + 'Event'} ])
```
fhcarter, line 909: oops

jcguion (approved):
LGTM

PR #417, opened 2023-09-29
Fix CamelComponent issue with structured messages via Google PubSub (at least)

### LINKED ISSUE #416: CamelComponent: structured messages from (at least) google pubsub fail
Reason is that we're using the object mapper to copy the map and it doesn't understand the types used in some of the google-specific headers.  When using structured message, we try & send all these values to Vantiq, and things go awry from there.

fhcarter:
Fixes #416

When dealing with (at least) message from google pubsub, the header values are not understood by the Jackson object mapper.  Instead, we'll revert to simply traversing the case insensitive structure given, converting it into a regular hashmap before sending it back to vantiq.  This is only an issue when asking for structured messages.


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/165/)

PR #418, opened 2023-10-02
Improve deserialization in Vantiq component producer

fhcarter:
(No issue reported)

Adds handling for getting byte arrays that really represent other things.  GooglePubSub seems to like to just pass things around as byte arrays. When we get a byte array, rather than only returning it as a byte array (since Vantiq, generally, can't do much with those), try a few ways to deserialize it. In the future, we may be able to use hints from media type (to figure out that, _e.g._, it's an image and do something other than sending a connector _notify),_ but for now (and when there's little/no information in the media type), try using Jackson to convert things, or, in the worst case, Java basic deserialization.

These extensions allow us to send & interpret message from Google Pubsub as a Vantiq system would send them (and in a manner that a Vantiq system can deal with them).

### DIFF
--- a/camelComponent/src/main/java/io/vantiq/extsrc/camel/VantiqConsumer.java
+++ b/camelComponent/src/main/java/io/vantiq/extsrc/camel/VantiqConsumer.java
@@ -98,7 +98,7 @@ public class VantiqConsumer extends DefaultConsumer {
             Map<String, Object> camelHdrs = null;
             Object camelBody = null;
             if (endpoint.isStructuredMessageHeader() && msgBody instanceof Map) {
-                camelHdrs = new HashMap<>();
+                camelHdrs = null;
                 Map<?,?> msgAsMap = (Map<?,?>) msgBody;
                 if (msgAsMap.get(STRUCTURED_MESSAGE_HEADERS_PROPERTY) instanceof Map) {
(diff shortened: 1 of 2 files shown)


fhcarter:
[Jenkins Run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/168/)

camelComponent/src/main/java/io/vantiq/extsrc/camel/VantiqConsumer.java
```diff
@@ -98,7 +98,7 @@ private void processMessage(Object msg, ExchangePattern pattern) {
             Map<String, Object> camelHdrs = null;
             Object camelBody = null;
             if (endpoint.isStructuredMessageHeader() && msgBody instanceof Map) {
-                camelHdrs = new HashMap<>();
```
fhcarter, line 101: If not provided, don't send anything -- rather than sending an empty map.

fhcarter:
[Jenkins Rerun after changes](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/169/) -- clean

fhcarter:
Bump

sfitts (approved):
LGTM

PR #429, opened 2023-10-31
Stream output from consumerOutputJson -- for Kinesis & other support

fhcarter:
Merge to staging branch.  No issue reported.

AWS Kinesis seems to function better when it gets StreamCache input.  As it turns out, this is what a `marshal` step  does (to Json via Jackson, but it's at a higher level so somewhat independent).  Since the purpose of our `consumerOutputJson` is to mimic the `marshal` step function (primarily for use in kamelets), we adopt this behavior.

However, since `consumerOutputJson` has been out there a bit (though probably very little use), we'll deprecate that and use `consumerOutputJsonStream` as the flag going forward.  The old one stays for now, still working the same way.  Backward compatibility and all that.

Add tests for same, verifying that the information is properly encoded in the `InputCacheStream` (or `ByteArrayInputStreamCache`) where appropriate.

### DIFF
--- a/camelComponent/src/test/java/io/vantiq/extsrc/camel/VantiqComponentTest.java
+++ b/camelComponent/src/test/java/io/vantiq/extsrc/camel/VantiqComponentTest.java
@@ -10,6 +10,7 @@ package io.vantiq.extsrc.camel;
 
 import static io.vantiq.extsrc.camel.VantiqEndpoint.ACCESS_TOKEN_PARAM;
 import static io.vantiq.extsrc.camel.VantiqEndpoint.CONSUMER_OUTPUT_JSON_PARAM;
+import static io.vantiq.extsrc.camel.VantiqEndpoint.CONSUMER_OUTPUT_JSON_STREAM_PARAM;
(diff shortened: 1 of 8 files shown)


camelComponent/src/test/java/io/vantiq/extsrc/camel/VantiqComponentTest.java
```diff
@@ -520,12 +550,21 @@ public void testVantiqConsumer() throws Exception {
     
     @Test
```
fhcarter, line 551: factor out common test code

camelComponent/src/test/java/io/vantiq/extsrc/camel/VantiqComponentTest.java
```diff
@@ -1058,13 +1123,24 @@ public void configure() {
 
                 from(vantiqJsonSenderUri)
                         .to(routeEndUri);
-
```
fhcarter, line 1061: new routes

fhcarter:
[Jenkins Run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/183/)

AlexBlumer (approved):
LGTM

PR #430, opened 2023-11-03
Use consumerOutputJsonStream header in generated assemblies

fhcarter:
No issue reported, merging to staging branch

Change `consumerOutputJson` EP option to `consumerOutputJsonStream`.

This conforms with the usage in the `VantiqConsumer` & better mimics the  _marshal/json..._ step we endeavor to replace (since it makes constructing assemblies much easier).


jcguion (approved):
LGTM

PR #439, opened 2024-01-26
Fix schema impedance mismatch kamelets vs. assemblies

fhcarter:
Kamelets allow parameters to be required AND have a default value.  Vantiq assembly configuration properties (generated from the Kamelet parameters) do not.  So, if we have something that's required but we find a default value, suppress the _required_ property -- the provision of the default value in the assembly will suffice.

Also fixed a few "it's groovy, not java" semicolon issues.

No issue reported

This is a PR into the staging branch -- which is likely to be PR'd/merged into master sometime next week or the week after.


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/197/)

jcguion (approved):
LGTM

PR #447, opened 2024-02-02
Initial Revision of Apache Camel Kamelet-based Assemblies

### LINKED ISSUE #387: camel: Add infrastructure for constructing assemblies

### LINKED ISSUE #390: camel: add sources to assemblies

### LINKED ISSUE #392: camel: create project

### LINKED ISSUE #419: camelComponent: Add equivalency definitions so that we can augment component headers with expected names
Many camel components provide headers that include common information, unfortunately using component specific header names.  For example, different components have headers that contain the content type (_e.g._, `image/jpeg`, `application/json`), but each component seems to provide this in a header named in a component specific manner. 

The Vantiq component can provide this information to/from Vantiq, but the Vantiq component cannot _know_ the "other end" of the Camel app to which it is connected, or the specific header name that that other component uses.

Instead, we propose offering a source configuration option that says to look for header XXX and add another header with that value before sending it back to Vantiq.  Or, augment the headers, adding XXX from YYY before sending to the camel app.  In this way, the creator of the connector source can define this behavior, and their Vantiq app can find these header values (_e.g._, contentType) in a place that they understand.

### LINKED ISSUE #423: camelConnector: Add ability to specify header duplication
That is, make use of #422.

This allows configuration of the source so that the associated Vantiq application need not know every possible component-specific header.  Which seems like a good thing:tm: for a integration system.

### LINKED ISSUE #444: Add Camel Assemblies doc/readme

fhcarter:
This is the merge of a long-running staging branch into master.  All individual changes have. been individually reviewed.

Fixes #387
Fixes #390 
Fixes #392 
Fixes #419
Fixes #423
Fixes #444
... and numerous unnamed issues

This PR adds support for the generation of Vantiq Assemblies from Apache Camel Kamelets.  These assemblies  each support moving information to or from (_sink_ or _source_, respectively) a specific external system via its Camel Component (the component is the extension point for Camel).  We currently do not save the output of the generation -- it appears in build/distributions as zip files, and can be uploaded to a Vantiq staging namespace for publication in a Vantiq Catalog.

This PR will upgrade our Camel support from 3.18 to 3.21.  Kamelet support was not in Camel 3.18, and kamelets are Camel version-specific.  (NB: The current Camel version is 4._something_, but I don't think there's any force driving us to move to that version.  OTOH, our Camel support here has probably not seen a great deal of use, so it could be a good time to upgrade -- sometime this year.)

Kamelets contain information about the route to use (which may include protocol transformations), parameters, as well as, potential, some descriptive or instructive information. 

The basic operation here is that there is gradle `buildsrc` code that takes the kamelet definition YAML file (which is in a jar file described in the dependencies), extracts the required information, and creates the associated Vantiq assembly.

Each assembly contains a source definition that includes (via reference) a route between the external system component and the Vantiq component, a Vantiq service (included appropriate _inbound_ or _outbound_ events), rules to move information between the service events and the source, and documents extracted from the kamelet definition. In some cases, there may be a (currently) _fred_ generated document as well, where there is some extra explanatory information required.  This is used when, for example, some external system operates on only strings or only binary, and we need to transform those into something that Vantiq can work with via the connector SDK.

The source is a CAMEL source (_i.e._, source implementation is CAMEL).

The assembly has configuration properties that are derived from the properties defined in the kamelet.  We use descriptors and/or format to determine if something is _secret_, and, if so, recommend that the associated value be placed in a Vantiq Secret, and referenced, as part of the installation, using the `@secrets()` notation.  We also make use of enum information present to construct assembly configurations properties as an Enum (added for 1.38, thanks @steve-langley).  This is particularly useful when there are long lists of items that must be correct (_e.g._, AWS regions) --- kamelets have this support so this is a simple transformation.

Once this is merged, some of the assemblies will be placed in a public Vantiq catalog for consumption by Vantiq folks or customers. For now (until otherwise decided), at least for _public_ public catalogs (_i.e._, those available on dev & api), we will publish only assemblies for which basic verification has been done.  Over time, for staging/test, we may put more and hope some other folks verify some.  In Apache Camel release 3.21, there are 160+ assemblies generated, which is more than I can test in any reasonable time frame.


camelComponent/README.md
```diff
@@ -1,7 +1,8 @@
 # Overview
```
fhcarter, line 1: There were changes made to the component to deal with things better. These are good improvements to the Vantiq component -- things we discovered using the generated assemblies (or support we required).

camelConnector/README.md
```diff
@@ -98,6 +98,7 @@ To change the logging settings, edit the logging config file
 which is an [Apache Log4j configuration file](https://logging.apache.org/log4j/2.x/manual/configuration.html). The logger 
 name for each class is the class's fully qualified class name, *e.g.* "io.vantiq.extjsdk.ExtensionWebSocketClient".
```
fhcarter, line 100: This fixes internal links to work :-)

jcguion (approved):
LGTM

camelAssemblies/README.md
```diff
@@ -0,0 +1,155 @@
...
+# Repository Contents
+
+Apache Camel, briefly, is organized into a set of _components_ that are used to integrate with other software 
+systems. The Vantiq Camel Connector provides the ability to load the components necessary to operate _routes_, 
+Camel's means for describing the interaction between information systems.
+
+Apache Kamelets are packages containing the routes and associated settings needed to set up connections using Apache
+Camel-K. 
+
+The assemblies contained in this repo take the information contained in the Kamelet definitions,
```
rblumer4, line 29: there are no assemblies "contained in this repo" at first glance.   
could you add a clarification here?   Maybe:
_This subproject will generate Vantiq assemblies from the camel definitions located <whereever they are>.  To generate the assembles, just run `../gradlew assemble`.  After that, the Vantiq assemblies will be located in zip files under the `build/distributions` directory. For example, build/distributions/ftp_sink_v3_21_0.zip_.

PR #458, opened 2024-03-06
Add support for route overrides -- and use one for the fair-sink

fhcarter:
It will occasionally be the case (hopefully not often) that we'll need to override the route generated from the kamelet.  This may be due to errors contained therein (which may just be because of our interactions) or may be cases where we'd like to enhance things.

The `AssemblyResourceGeneration` task now looks in `resources/routeOverrides` and will use the route there when present in lieu of the generated version.

For the FHIR sink, we find that one of the endpoint parameters does not appear to be accepted, so we'll remove it.  Also, since this _sink_ is really a request/response thing, we add a Vantiq Note that describes how to make use of that.  Using the usual PUBLISH to SERVICE EVENT mechanism, you can successfully send operations (once you decode their spec!), but a _write only_ operation for, for example, a search, isn't terribly useful.  So we add to the route to return the result.  And the note describes how to get at that.  Unfortunately, this requires working against the underlying source (which is somewhat ghostly since 1) sources can't be visible, but 2) they show up in the _active resource control_ and in the drop downs in the Vail editor.

Also, as a matter of convenience to me, the `build.gradle` file has ordering directives relating _assemble_, _importAssemblies_, and _publishAssemblies_.  Just makes it easier locally to _fire & forget_.

### DIFF
--- a/camelAssemblies/build.gradle
+++ b/camelAssemblies/build.gradle
@@ -48,21 +48,22 @@ zipAssemblies.mustRunAfter( 'generateAssemblyResources' )
 
 tasks.register('importAssemblies') {
     outputs.upToDateWhen { false }
-    def importCount = 0
-    def skipCount = 0
-    String catalogProfile = project.rootProject.findProperty('camelAssembliesProfile') as String
-    logger.lifecycle('Found profile: {}', catalogProfile as String)
-    if (!catalogProfile?.trim()) {
-        throw new GradleException('Project property "camelAssembliesProfile" was missing or empty')
-    }
-
-    def catalogProjectList = project.rootProject.findProperty("${catalogProfile}_camelAssembliesList")
-    def catalogVantiqCommand = project.rootProject.findProperty("${catalogProfile}_camelAssembliesVantiq")
-    if (!catalogVantiqCommand?.trim()) {
-        catalogVantiqCommand = 'vantiq'
-        logger.lifecycle('Using {} as the "vantiq" command', catalogVantiqCommand)
-    }
     doLast {
+        def importCount = 0
+        def skipCount = 0
+        String catalogProfile = project.rootProject.findProperty('camelAssembliesProfile') as String
+        logger.debug('Found profile: {}', catalogProfile as String)
+        if (!catalogProfile?.trim()) {
+            throw new GradleException('Project property "camelAssembliesProfile" was missing or empty')
+        }
+
+        def catalogProjectList = project.rootProject.findProperty("${catalogProfile}_camelAssembliesList")
+        def catalogVantiqCommand = project.rootProject.findProperty("${catalogProfile}_camelAssembliesVantiq")
+        if (!catalogVantiqCommand?.trim()) {
+            catalogVantiqCommand = 'vantiq'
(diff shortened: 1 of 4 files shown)


camelAssemblies/build.gradle
```diff
@@ -162,3 +162,6 @@ tasks.register('publishAssemblies') {
     }
 }
```
fhcarter, line 164: Allow a `gradle assemble importAssemblies publishAssemblies` command.  Takes a while for the fully bunch, so nice to fire & forget.

camelAssemblies/src/main/resources/routeOverrides/fhir_sink/v3_21_0/fhir_sink.routes.yaml
```diff
@@ -0,0 +1,43 @@
...
+                                    pretty-print: '{{prettyPrint}}'
+                    -   simple: ${properties:encoding} =~ 'XML'
+                        steps:
+                        -   unmarshal:
+                                fhirXml:
+                                    fhir-version: '{{fhirVersion}}'
+                                    pretty-print: '{{prettyPrint}}'
+            -   to:
+                    uri: fhir://{{apiName}}/{{methodName}}
+                    parameters:
```
fhcarter, line 23: `inBody: "resource"` was in this list, but the underlying component doesn't seem to support it (blows up).  So we'll remove it.

camelAssemblies/src/main/resources/routeOverrides/fhir_sink/v3_21_0/fhir_sink.routes.yaml
```diff
@@ -0,0 +1,43 @@
...
+                        log: '{{log}}'
+                        prettyPrint: '{{prettyPrint}}'
+                        lazyStartProducer: '{{lazyStartProducer}}'
+                        proxyHost: '{{?proxyHost}}'
+                        proxyPassword: '{{?proxyPassword}}'
+                        proxyPort: '{{?proxyPort}}'
+                        proxyUser: '{{?proxyUser}}'
+                        accessToken: '{{?accessToken}}'
+                        username: '{{?username}}'
+                        password: '{{?password}}'
```
fhcarter, line 36: the `marshal` & `to` below send result data back to Vantiq.  Useful in a SELECT, ignored in a PUBLISH.

jcguion (approved):
LGTM

camelAssemblies/src/main/resources/vantiqNotes/fhir_sink.md
```diff
@@ -0,0 +1,239 @@
...
+
+var msg = { resourceType: resourceType }
+var headers = {CamelFhir.url = urlParam }
+
+publish {headers: headers, message: msg } to SERVICE EVENT  "com.vantiq.extsrc.camel.kamelets.v3_21_0.fhir_sink.fhir_sink_service/fhir_sink_serviceEvent"
+
+return null
+```
+
+And this can be invoked using "Patient" as the resourceType, and "Patient?name=smith".
```
rblumer4, line 33: typo:  missing text at end, like "as the urlParam"

PR #466, opened 2024-03-23
Update pypi metadata to reflect current dependencies

fhcarter:
Really only supporting 3.10, so make that clear.  Update versions  for publication


AlexBlumer (approved):
LGTM

PR #482, opened 2024-04-23
Fix DockerFile for camelconnector to fix newly reclassified to HIGH security issues

fhcarter:
Fix quay.io security warnings.

Quite a number had crept in since last week's publication (or, more correctly, I think they were reclassified).

Anyway, add package upgrades for `libssl3` & `libcrypto3` to Dockerfile to correct these.

Upgrade version of resulting docker image to camel connector assembly defaults. (Skipping one patch number since that's the one with large number of HIGH security issues.)

Found when publishing new version.  Tracked & fixed, then published under unused tag to check.  Result:  Only 1 LOW warning (due to Camel version).


fhcarter:
(no Jenkins -- nothing to test)

mer2github (approved):
LGTM

PR #487, opened 2024-05-06
Upgrade jinja2

fhcarter:
Fix dependabot security warning.

(Done in two steps so we can publish extpsdk to pypi from master.  Just a formality, but...)


mer2github (approved):
LGTM

PR #491, opened 2024-06-06
Fix requests-related CVE

### LINKED ISSUE #489: Extpsdk: Fix dependabot warning re: requests < 2.32.0

fhcarter:
Fixes #489

Upgrade requests to newer version.  Deal with CVE.

Will be parallel change for python exec source.  Two-stepper so that we can push to pypi from master.


sfitts (approved):
LGTM

PR #510, opened 2024-08-16
Upgrade code & environment for JDK 17.  But not using it yet.

fhcarter:
Upgrade source to JDK 17 level without warnings.

In prep for some other work to be done soon, upgrade things to run with JDK 17.  This is mostly changes to use of the Java security manager (which is now deprecated) that we were using in tests.  Replaced with try-finally & a so-called `ExitProcessor` that allows us to through exceptions rather than exit (useful in testing).  This is a test-only change.

Upgraded/changed Lombok plugin since the one in use is in maintenance-only mode and doesn't support Java 17.

Other minor tweaks.

### DIFF
--- a/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVMain.java
+++ b/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVMain.java
@@ -130,14 +130,14 @@ public class CSVMain {
             }
         }
         if (exitProcessor != null) {
-            exitProcessor.doExit(code);
+            exitProcessor.processExit(code);
         } else {
             System.exit(code);
         }
     }
     
-    interface ExitProcessor {
-        default void doExit(int code) {
+    public interface ExitProcessor {
+        default void processExit(int code) {
             System.exit(code);
         }
     }
--- a/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVMain.java
+++ b/CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVMain.java
@@ -41,6 +41,8 @@ public class CSVMain {
     static final int NO_AUTH_EXIT = 1;
     static final int NO_SOURCE_EXIT = 2;
     static final int NO_SERVER_EXIT = 3;
+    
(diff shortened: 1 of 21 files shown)


CSVSource/src/main/java/io/vantiq/extsrc/CSVSource/CSVMain.java
```diff
@@ -41,6 +41,8 @@ public class CSVMain {
     static final int NO_AUTH_EXIT = 1;
     static final int NO_SOURCE_EXIT = 2;
     static final int NO_SERVER_EXIT = 3;
+
```
fhcarter, line 44: this is a common pattern.  It's seen in many of the connectors that used the security manager

buildSrc/build.gradle
```diff
@@ -1,7 +1,6 @@
-
 plugins {
     id 'java'
```
fhcarter, line 2: old Lombok processor no longer maintained, and doesn't work with JDK 17.  Upgrade to IntelliJ recommended one.  Works fine w/o code changes (other than build.gradle code)

buildSrc/build.gradle
```diff
@@ -12,7 +11,7 @@ ext {
     apacheCommonsLangVersion = '3.12.0'
     jacksonVersion = '2.14.2'
     log4JVersion='2.17.2'
```
fhcarter, line 13: use common version throughout.

jdbcSource/build.gradle
```diff
@@ -1,10 +1,14 @@
-group 'io.vantiq'
-version 'unspecified'
```
fhcarter, line 1: Using new-style plugins closure requires moving it in gradle file.  No substantive changes here other than the new Lombok plugin

fhcarter:
[Jenkins run ](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/256/)

fhcarter:
(Changes tested using Javas 11 & 17.  17 required for a separate upgrade (to be commenced soon))

mer2github (approved):
LGTM

PR #524, opened 2024-09-17
Improve ObjectRec test reliability & consistency

### LINKED ISSUE #523: ObjRec: Add single test that verifies that some camera is valid
As of #522, we'll skip tests where there's no camera found.  

Add one test that doesn't do this -- that is, it will fail if no valid camera is found.  This test will have an obvious name so we know why it's failing, but we don't have to generate bunch of junk failures.

fhcarter:
Fixes #523 

Adds a test (and some minor refactoring) that will fail if no valid IP cameras are found.  This avoids tons of result pollution when such things happen, but still lets us know without have to count skipped tests.

Also, generated new RTSP camera & re-enabled RTSP-based network retriever test.  We'll see how long this one lasts.

### DIFF
--- a/objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/imageRetriever/TestNetworkStreamRetriever.java
+++ b/objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/imageRetriever/TestNetworkStreamRetriever.java
@@ -9,7 +9,7 @@ import java.util.Map;
 import io.vantiq.extsrc.objectRecognition.ObjRecTestBase;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+
 import org.junit.Test;
 
 import io.vantiq.extsrc.objectRecognition.NoSendORCore;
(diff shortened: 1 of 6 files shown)


objectRecognitionSource/src/test/java/io/vantiq/extsrc/objectRecognition/imageRetriever/TestNetworkStreamRetriever.java
```diff
@@ -9,15 +9,17 @@
...
-import org.junit.Ignore;
+
 import org.junit.Test;
 
 import io.vantiq.extsrc.objectRecognition.NoSendORCore;
 import io.vantiq.extsrc.objectRecognition.exception.ImageAcquisitionException;
 
 public class TestNetworkStreamRetriever extends ObjRecTestBase {
 
-    static final String RTSP_CAMERA_ADDRESS = "rtsp://wowzaec2demo.streamlock.net/vod/mp4:BigBuckBunny_115k.mp4";
```
fhcarter, line 20: Defined new camera (free account) to validate RTSP network stream retriever...

sfitts (approved):
LGTM

fhcarter:
Updated most of the TestYolo* tests.  They were using (and failing, often) one of the network cameras we picked.  Turns out that that some of these come & go unreliably. 

Changed these tests to run based on a file rather than on an IP camera.  What they were testing had nothing to do with the camera, so this just improves reliability of the tests.

Changed the TestNetworkRetriever tests to fetch multiple images from the IP camera & RTSP camera, and verify that all the images are different (using hash digest).  By doing so, we better verify that our IP camera based sources work, and we exercise that code a bit more (since it's not longer being exercised by the TestYolo tests -- which ran it but didn't care about the results).

PR #526, opened 2024-09-19
Fix some camel deprecation warnings

fhcarter:
No issue reported

Camel's yaml format now _encourages_ camelCase rather than _kebab-case_.  Fix places in tests & route overrides where this happens.

There are still a couple of places where this shows up, but those are preserved from the Camel Kamelet definitions, so we'll leave those at the source kamelets define them.  The kebab-case properties still work;  we just get deprecation warnings in the code.  Where we're generating these, we fix them.

### DIFF
--- a/buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
+++ b/buildSrc/src/main/groovy/io/vantiq/extsrc/assy/tasks/AssemblyResourceGeneration.groovy
@@ -73,7 +73,7 @@ class AssemblyResourceGeneration extends DefaultTask {
 
     public static final String FROM_ROUTE_TAG = 'from'
     public static final String ID_ROUTE_TAG = 'id'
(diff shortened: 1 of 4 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/275/) -- underway.

sfitts:
> kebab-case

That's a new one for me.  Guess it is hard to come up with another animal to match with camel and snake.  Also, it only makes sense that Camel should prefer camel casing ;).

sfitts (approved):
LGTM

fhcarter:
no argument:  Just using their words...

>WARN  The kebab-case 'route-template' is deprecated and it will be removed in the next version. Use the camelCase 'routeTemplate' instead.

fhcarter:
At least [one definition](https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/) says _snake\_case_ vs. _kebab-case_ is **_** vs **-**

PR #530, opened 2024-10-25
EXTPSDK: Add ability to provide keyword arguments to connect() call

### LINKED ISSUE #528: extpsdk:  Add ability to add connect arguments (as was done w/base python sdk)

fhcarter:
Fixes #528

It is occasionally necessary to provide extra arguments to the websockets connect call.  This is typically a case for disabling SSL verification in a development environment.

To enable this, we provide for the `connectKWArgs` property to be part of the `server.config` file used to specify the connection to Vantiq in a connector.  The content of this property should be a valid JSON string specifying the names & values of the keyword arguments.  Note that this is "just" a string -- making Python calls is not supported.  A typical case will be something like

```
connectKWArgs={ "ssl": false}
```

as part of the `server.config` file.

### DIFF
--- a/extpsdk/README.md
+++ b/extpsdk/README.md
@@ -82,6 +82,8 @@ In addition to the configuration properties shown above, the `server.config` fil
 
 *   `sendPings`: A boolean property that, if set to `true`, enables the SDK to send ping messages to the Vantiq Server. 
 The ping messages are handled by the underlying websockets library.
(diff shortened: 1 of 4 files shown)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/281)

jcguion (approved):
LGTM

PR #531, opened 2024-10-29
Disable SSL processing in a manner usable by the Python exec source connector

### LINKED ISSUE #528: extpsdk:  Add ability to add connect arguments (as was done w/base python sdk)

fhcarter:
Fixes #528 (for real!)

Previously, we added the ability to handle keyword arguments, passing them through to the Vantiq websocket connection.

Unfortunately, when making use of this in the connector, there are inconsistencies between the keyword arguments supported by the "regular" Vantiq connection of this SDK's websocket connection.  

To handle this, we extend the keyword argument support in include `disableSslVerification=true`.  When this is provided, the connector SDK builds a special SSL context with hostname & certification verification disabled, and uses that as the keyword `dict`-ionary.  This is acceptable to both the web sockets connection & the Vantiq rest connection.

Updated a few versions & documentation...

### DIFF
--- a/extpsdk/build.gradle
+++ b/extpsdk/build.gradle
@@ -11,7 +11,7 @@ ext {
         pypi_repo = 'testpypi'
     }
     if (!project.rootProject.hasProperty('vantiqConnectorSdkVersion')) {
-        vantiqConnectorSdkVersion = '1.3.0'
+        vantiqConnectorSdkVersion = '1.3.1'
     }
 }
 
--- a/extpsdk/build.gradle
+++ b/extpsdk/build.gradle
@@ -11,7 +11,7 @@ ext {
         pypi_repo = 'testpypi'
     }
     if (!project.rootProject.hasProperty('vantiqConnectorSdkVersion')) {
-        vantiqConnectorSdkVersion = '1.3.1'
+        vantiqConnectorSdkVersion = '1.3.2'
     }
 }
 
--- a/extpsdk/build.gradle
+++ b/extpsdk/build.gradle
@@ -11,7 +11,7 @@ ext {
         pypi_repo = 'testpypi'
     }
     if (!project.rootProject.hasProperty('vantiqConnectorSdkVersion')) {
-        vantiqConnectorSdkVersion = '1.3.2'
+        vantiqConnectorSdkVersion = '1.3.3'
     }
 }
 
--- a/extpsdk/pytest.ini
+++ b/extpsdk/pytest.ini
@@ -5,7 +5,7 @@ filterwarnings =
 junit_logging = all
 junit_suite_name=io.vantiq.extsrc.extpsdk
 minversion = 6.0
-addopts = -rE -q --junit-xml=./build/test-results/test/TEST-io.vantiq.extsrc.extpsdk.pytest.xml
+addopts = -rE -q --html=./build/reports/tests/test/index.html --junit-xml=./build/test-results/test/TEST-io.vantiq.extsrc.extpsdk.pytest.xml
 asyncio_mode = strict
 testpaths = src/test/python
 pythonpath = src/main/python
--- a/extpsdk/src/main/python/vantiqconnectorsdk.py
+++ b/extpsdk/src/main/python/vantiqconnectorsdk.py
@@ -226,7 +226,10 @@ class VantiqSourceConnection:
                     disable_ssl = kw_temp.get(VantiqConnector.DISABLE_SSL_VERIFICATION)
                     if disable_ssl is not None and disable_ssl:
                         self.disable_ssl_check = True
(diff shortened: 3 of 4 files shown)


extpsdk/build.gradle
```diff
@@ -11,7 +11,7 @@ ext {
         pypi_repo = 'testpypi'
     }
     if (!project.rootProject.hasProperty('vantiqConnectorSdkVersion')) {
-        vantiqConnectorSdkVersion = '1.3.0'
```
fhcarter, line 14: took some tries to get it all worked out.

Intervening versions put only in test-pypi, so no interference.

extpsdk/pytest.ini
```diff
@@ -5,7 +5,7 @@ filterwarnings =
 junit_logging = all
 junit_suite_name=io.vantiq.extsrc.extpsdk
 minversion = 6.0
-addopts = -rE -q --junit-xml=./build/test-results/test/TEST-io.vantiq.extsrc.extpsdk.pytest.xml
```
fhcarter, line 8: tired of parsing XML, build html report

extpsdk/src/main/python/vantiqconnectorsdk.py
```diff
@@ -220,7 +223,15 @@ def __init__(self, source_name: string, config: Union[dict, None]):
             if kwArgString is not None:
                 try:
                     kw_temp : dict = json.loads(kwArgString)
-                    self.connect_kw_args = kw_temp
```
fhcarter, line 223: when we get this special case, construct an SSL context to be used for the connection.  

`ssl=false` disables ssl (so our server refuses to connect), and I don't see a way to do this in pure JSON.  So we special case for this a bit.

fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/282/)

jcguion (approved):
LGTM

fhcarter:
Note: the underlying issue here is that for our Vantiq/REST connection, `{ssl=false}` disables SSL verification.  However, for our Websockets connection, `{ssl=false}` tells it make the connection with SSL disabled.  Which cause the Vantiq server to reject the connection.

So a bit of sleight of hand was required...

extpsdk/pytest.ini
rblumer4: @fhcarter, does this get generated by our Jenkins job, too?
Do we need to add a step to upload that report to Jenkins?

extpsdk/src/main/python/vantiqconnectorsdk.py
```diff
@@ -220,7 +223,15 @@ def __init__(self, source_name: string, config: Union[dict, None]):
...
+                    disable_ssl = kw_temp.get(VantiqConnector.DISABLE_SSL_VERIFICATION)
+                    if disable_ssl is not None and disable_ssl:
+                        self.disable_ssl_check = True
+                        ctx = ssl.create_default_context()
+                        ctx.check_hostname = False
+                        ctx.verify_mode = ssl.CERT_NONE
+                        self.connect_kw_args = {"ssl": ctx}
```
rblumer4, line 232: is there any use case for merging this SSL setting with other settings that might be in kw_temp (minus the DISABLE_SSL_VERIFICATION setting) ?

fhcarter: We could do that.  It doesn't look like the others, but having it available isn't a bad thing.  So far, I just rerun locally and find issues that way, but having it around wouldn't hurt.

extpsdk/pytest.ini
fhcarter: We could do that. It doesn't look like the others, but having it available isn't a bad thing. So far, I just rerun locally and find issues that way, but having it around wouldn't hurt.

extpsdk/src/main/python/vantiqconnectorsdk.py
fhcarter: There may be.  But I didn't want to deal with conflict resolution, etc., at this point, so I figured we could wait and see if there's any demand.  This is the first time this has come up, and it's primarily, in that case, an issue with the client environment.  But it's useful for dealing with installations with self-signed certs, so it seemed worthy of adding.

extpsdk/pytest.ini
rblumer4: OK, I added a publishHTML for that report to both Jenkins build definitions.
```
            // publish the generated python-sdk HTML report  (see PR 531)
            publishHTML target: [
                    allowMissing: false,
                    alwaysLinkToLastBuild: false,
                    keepAll: true,
                    reportDir: 'vantiq-extension-sources/extpsdk/build/reports',
                    reportFiles: 'tests/test/index.html',
                    reportTitles: 'ExtPSDK Report',
                    reportName: 'External-Python SDK Report'
            ]
```

PR #534, opened 2024-11-07
Protect .onlyIf & .dependsOn from cases where tasks don't exist

### LINKED ISSUE #533: Fix build failure when dockerRegistryUser/Password not set
When these are not set, the build*Image task isn't build.  Then, the `.onlyIf` on those tasks crashes the build since the task doesn't exist.

Need protection around those when task doesn't exist.

fhcarter:
Fixes #533

In cases where users are building, they may not have the various docker* gradle properties set.  When that's the case, the docker/dcompose plugins don't build those tasks, and gradle gets upset.

These changes validate that the tasks exist before setting the onlyIf, etc., values.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,9 +110,7 @@ subprojects {
             }
         }
     }
-    if (project.tasks.findByName('buildConnectorImage')) {
-        buildConnectorImage.dependsOn assemble
-    }
+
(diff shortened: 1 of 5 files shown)


jcguion (approved):
LGTM

PR #538, opened 2024-11-20
EXTPSDK: Update aiohttp as per dependabot reports

fhcarter:
As with Vantiq Python SDK, update requirements.  Update version number accordingly

Also fix .md type in PES.

Similar update to PES coming once this is approved, merged, and pushed.  We do this so we're always pushing from master.


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/293)

PR #542, opened 2025-01-21
Set up FHIR server for testing

fhcarter:
Merge into staging branch for the FHIR assembly work.

No issue reported.

Found a useful FHIR server that has a docker image.  From the gradle file, run said server & load some data.  We don't load a lot since we're not testing FHIR -- just enough that we can fetch/update/delete records.

The server we start (open source, ASL) is configured in-memory only, so no data survives the test suite.  This allows us to muck with it at will.

Added some docker-compose & associated configurations to allow developers to have server with longer term storage and/or just use docker compose to start things.  Sometimes more convenient than running via gradle each time.


fhirAssembly/build.gradle
```diff
@@ -0,0 +1,213 @@
+plugins {
```
fhcarter, line 1: New build file for new subproject

fhirAssembly/build.gradle
```diff
@@ -0,0 +1,213 @@
...
+
+    // Used for uploading documents (images) to VANTIQ with VANTIQ SDK
+    testImplementation "io.vantiq:vantiq-sdk:${vantiqSDKVersion}"
+
+    // Use JUnit test framework
+    testImplementation 'junit:junit:4.13.1'
+    // Data to preload into our FHIR server
+    testResources "vantiq.testArtifacts:fhirR4Sample:1.0@zip"
+}
+
```
fhcarter, line 70: We've found some sample data in FHIR R4 form.  We've taken a small subset of this & include this as the dependency above.  The `unzip...` and `upload...` tasks load this data into our server.  Since the server is configured as in-memory only, we do this at startup.

fhirAssembly/build.gradle
```diff
@@ -0,0 +1,213 @@
...
+                    req.setRequestProperty('Content-Type', 'application/json; charset=UTF-8')
+                    // Note:  If accept header isn't set (or not set to app/json or app/xml), you get HTML back,
+                    // and the work isn't actually done.  Only took a day to figure that out :-(
+                    req.setRequestProperty('Accept', 'application/json')
+                    req.setDoOutput(true)
+                    req.setDoInput(true)
+                    req.getOutputStream().write(sampleFile.bytes)
+                    logger.debug('Uploading to {}: {}', redirectedServer, new String(sampleFile.bytes))
+                    logger.info "Status code: ${req.getResponseCode()}" // HTTP request done on first read
+                    switch (req.getResponseCode()) {
```
fhcarter, line 115: Have to handle redirects here due to the `setFixedLengthStreamingMode()` call.  At least one of the samples is long enough that it needs to be chunked, so...

fhirAssembly/build.gradle
```diff
@@ -0,0 +1,213 @@
...
+        image = 'hapiproject/hapi:latest'
+        exposedPorts = [ '8080' ]
+        portBindings = ['8090:8080']
+        env = [
+            'HAPI_FHIR_SERVER_ADDRESS=http://localhost:8090/fhir'
+        ]
+    }
+}
+
+/**
```
fhcarter, line 173: The server takes a while to start.  So, we'll repeatedly try & fetch the metadata/conformance statement.  This isn't available until to server is ready to go, so it suffices as a a ready check.  We have a configurable max time to wait, & we try to sample it periodically without sucking up the whole machine.  Generally, it seems to take 20-50 seconds.

settings.gradle
```diff
@@ -12,6 +12,8 @@ include 'testConnector'
 include 'objectRecognitionSource'
 include 'jdbcSource'
 include 'jmsSource'
+include 'jmsSource'
+include 'fhirAssembly'
```
fhcarter, line 15: add new subproject...

fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/297/)

sfitts (approved):
LGTM

fhirAssembly/build.gradle
```diff
@@ -0,0 +1,213 @@
+plugins {
+    id 'java'
+    id("io.freefair.lombok") // version "6.3.0" // version "8.7.1"
```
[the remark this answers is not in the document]
sfitts: Cool -- probably should update the ag2rs repo to use this as well  (not that we'll be able to move to Java 17 anytime soon).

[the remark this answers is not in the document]
fhcarter: I don't think I did any in-depth research,  just found one that worked for what we needed.  If there's a better one, let me know.  I think the version is a bit out of date.  I think there may have been a java or gradle version issue, which is why it stepped around a bit.  Now that we're on gradle 8, I think its version should match the gradle version.  But not a high priority since it works...

PR #577, opened 2025-04-08
Remove OPCUA server that no longer exists

fhcarter:
The OPCUA tests have a bunch of public servers they test against.  One appears to have had its node evaporated, so remove it.


jcguion (approved):
LGTM

PR #580, opened 2025-04-11
Add description to NLP assembly

fhcarter:
No issue reported.

Add a description to the assembly:

- This assembly adds conversational language capability to the namespace. This capability works in conjunction with an external natural language interpreter to understand an utterance. The com.vantiq.nlp.InterpretCLU component enables this in an event handler.

This description & description of the included component will be display in the component catalog discovery pane.  If there is formatting capability for the description, I can tweak the component name to be `code` looking.  But I don't know that that is present...

(Question: Do we have generic name for the catalog?  In docs (_e.g_, the release notes), how do we talk about said catalog?)


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/330/)

rblumer4:
> (Question: Do we have generic name for the catalog? In docs (_e.g_, the release notes), how do we talk about said catalog?)

Not yet.  On staging, I called it _VantiqComponentCatalog_, so we could call it "_the component catalog_" or "_the Vantiq component catalog_".  But we could name it something different on dev.US  (and then rename staging to match).

AlexBlumer (approved):
LGTM

PR #585, opened 2025-04-24
Update NLP Assembly to be an assembly...

fhcarter:
Guess our Vail code that uses things is less constrained than the UI.  So I can install something that's not an assembly from a catalog in the tests, but the UI doesn't seem to like that configuration for installation purposes (quite correctly).

Update the project to declare its assembly-ness.  Hopefully, that'll make things better...


fhcarter:
[Jenkins run](https://jenkins.vantiq.com/jenkins/job/vantiq-extension-sources-branch/335/)

rblumer4 (approved):
LGTM

PR #591, opened 2026-04-17
fix single-quote typo on python toml template

rblumer4:
This will fix the "Homepage" link on the Vantiq Python Execution Connector page at pypi.org


AlexBlumer (approved):
LGTM

PR #592, opened 2026-04-20
Update the published versions for Python SDK and Python Exec Connector

AlexBlumer:
Didn't actually do this in the previous PR.

Also published the new versions to PyPi


mer2github (approved):
LGTM

PR #594, opened 2026-04-22
Fix test in OPCUA source

### LINKED ISSUE #593: One of the OPCUA tests fails when run in k8s
The tests wants to make the creation of a directory fail. It did so in Linux by trying to make a new top-level directory which is generally illegal. However, something in the configuration of a k8s installation made it unexpectedly legal.

AlexBlumer:
Fixes #593


skyler1ackerman (approved):
LGTM

PR #598, opened 2026-09-08
Fix failing OPCUA test

AlexBlumer:
No issue filed. One of the servers is now failing for the "None" security tests.


sfitts (approved):
LGTM

